PR #2, opened 2020-11-06
WIP: SSTOOLKIT-14: login token w/o response parsing ...

unserializable:
For comments, not merging at this point.

The operation itself is such that it will only succeed when token is NOT already logged in, so handling such situations where 'no operation is needed' but error is returned warrants some consideration for the whole toolkit

For this particular call, response is with HTTP status code 409 when token is already logged in, body
`` {"status":409,"error":{"code":"action_not_possible"}}`

### DIFF
--- a/sstoolkit/controllers/base.py
+++ b/sstoolkit/controllers/base.py
@@ -8,6 +8,8 @@ from ..core.version import get_version
 from sstoolkit.configuration.configuration import Configuration
 from sstoolkit.api_client.api_client import ApiClient
 from sstoolkit.api.clients_api import ClientsApi
+from sstoolkit.api.tokens_api import TokensApi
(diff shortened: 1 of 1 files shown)


bertvi:
I see that you take configuration data from the command line, but I think at this point it should be taken from configuration files, like for example the url and api-key. 

Please add some unit tests aswell. 

Otherwise looks OK.

bertvi (changes requested):
I see that you take configuration data from the command line, but I think at this point it should be taken from configuration files, like for example the url and api-key.

Please add some unit tests aswell.

Otherwise looks OK.

unserializable:
Abandoning, as work done in local branches and PR3 renders this outdated.

PR #4, opened 2020-11-13
Controllery sstoolkit 14

unserializable:
Some PR3 tests are borken here and tests for token operations need to be added, but mainly seems to work, please verify.

General idea here is:
  * have base controller
  * have subcommands inheriting from base controller and able to define their own arguments

Token operations added are 'list' and 'login'

### DIFF
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -12,6 +12,7 @@ security-server:
     owner_member_class: <MEMBER_CLASS>
     owner_member_code: <MEMBER_CODE>
     security_server_code: <SERVER_CODE>
+    software_token_id: 0 # default software token ID is zero
     software_token_pin: <SOFT_TOKEN_PIN>
 
   - name: <SECURITY_SERVER_NAME>
@@ -21,5 +22,6 @@ security-server:
     owner_member_class: <MEMBER_CLASS>
     owner_member_code: <MEMBER_CODE>
     security_server_code: <SERVER_CODE>
+    software_token_id: 0 # default software token ID is zero
     software_token_pin: <SOFT_TOKEN_PIN>
 
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -12,7 +12,7 @@ security-server:
     owner_member_class: <MEMBER_CLASS>
     owner_member_code: <MEMBER_CODE>
     security_server_code: <SERVER_CODE>
-    software_token_id: 0 # default software token ID is zero
+    software_token_id: 0 # <SOFT_TOKEN_ID> -- default software token ID is zero
     software_token_pin: <SOFT_TOKEN_PIN>
 
   - name: <SECURITY_SERVER_NAME>
@@ -22,6 +22,6 @@ security-server:
     owner_member_class: <MEMBER_CLASS>
     owner_member_code: <MEMBER_CODE>
     security_server_code: <SERVER_CODE>
-    software_token_id: 0 # default software token ID is zero
+    software_token_id: 0 # <SOFT_TOKEN_ID> -- default software token ID is zero
     software_token_pin: <SOFT_TOKEN_PIN>
 
--- a/tests/unit/test_token.py
+++ b/tests/unit/test_token.py
@@ -2,9 +2,9 @@ import unittest
 import urllib3
 from xrdsst.controllers.token import TokenController
 
-class TestInit(unittest.TestCase):
+class TestToken(unittest.TestCase):
     ss_config = {
-        'logging': [{'file': '/tmp/xrdsst_test_init_log', 'level': 'INFO'}],
+        'logging': [{'file': '/tmp/xrdsst_test_token_log', 'level': 'INFO'}],
         'security-server':
             [{'name': 'ssX',
               'url': 'https://192.168.1.1:8999/api/v1',
--- a/tests/unit/test_token.py
+++ b/tests/unit/test_token.py
@@ -1,13 +1,78 @@
 import unittest
+from datetime import datetime
+from unittest import mock
+
+from dateutil.tz import tzutc
+
+from xrdsst.models import TokenStatus, TokenType, Key, KeyUsageType, TokenCertificate, CertificateOcspStatus, \
+    CertificateStatus, CertificateDetails, KeyUsage
+from xrdsst.models.token import Token
+
 import urllib3
 from xrdsst.controllers.token import TokenController
+
+class TokenTestData:
+    # JSON of the response unusable due to assertions in generated Client API, define response in Python
+    token_login_response = Token(
+        available=True,
+        id=0,
+        logged_in=True,
+        name='softToken-0',
+        possible_actions=None,
+        read_only=False,
+        saved_to_configuration=True,
+        serial_number=None,
+        status=TokenStatus.OK,
+        token_infos=[{'key': 'Type'}, {'value': 'Software'}],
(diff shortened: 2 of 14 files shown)


bertvi:
> Some PR3 tests are borken here and tests for token operations need to be added, but mainly seems to work, please verify.
> 
> General idea here is:
> 
> * have base controller
> * have subcommands inheriting from base controller and able to define their own arguments
> 
> Token operations added are 'list' and 'login'

The idea of creating a base controller to have some common methods for all the controllers is very good.

bertvi:
When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have created "resources" as a normal folder, not as a module. I tried creating it as a module and then it works. 

Runngin "xrdsst init" seems to work fine. Then after that running "xrdsst token list" seems also to work find as it lists tokens information, but finally, when running "xrdsst token login", I get the following error: "KeyError: software_token_id".
pr

unserializable:
> When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have information, but finally, when running "xrdsst token login", I get the following error: "KeyError: software_token_id".

Ay, 'software_token_id' is not defined in your base.yaml -- since it is also missing from the template (I only added it locally).

unserializable:
> When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have created "resources" as a normal folder, not as a module. I tried creating it as a module and then it works.

I wonder what in our environments is causing this, it works for me both in PyCharm and on command line (virtualenv). Can you tell me how you were running ``xrdsst`` when getting that message, to reproduce?

bertvi:
> > When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have created "resources" as a normal folder, not as a module. I tried creating it as a module and then it works.
> 
> I wonder what in our environments is causing this, it works for me both in PyCharm and on command line (virtualenv). Can you tell me how you were running `xrdsst` when getting that message, to reproduce?

I installed the xrdsst with "python setup.py install", then filled in the cofiguration file with the configuration data of my local security servers and then ran "xrdsst init". After getting the error, I tried creating a new module named "res" and copied the resource under that module and updated the files using the module and it worked.
a

unserializable:
> > When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have information, but finally, when running "xrdsst token login", I get the following error: "KeyError: software_token_id".
> 
> Ay, 'software_token_id' is not defined in your base.yaml -- since it is also missing from the template (I only added it locally).

Addded in commit f0ae6471bc281b251f8863fad47d80cc6566b74c

unserializable:
> > > When running the toolkit, I first get the error that module "xrdsst.resources" is not found. Here the problem is that you have created "resources" as a normal folder, not as a module. I tried creating it as a module and then it works.
> > 
> > 
> > I wonder what in our environments is causing this, it works for me both in PyCharm and on command line (virtualenv). Can you tell me how you were running `xrdsst` when getting that message, to reproduce?
> 
> I installed the xrdsst with "python setup.py install", then filled in the cofiguration file with the configuration data of my local security servers and then ran "xrdsst init". After getting the error, I tried creating a new module named "res" and copied the resource under that module and updated the files using the module and it worked.
> a

Ok, your Python is younger than mine. The import form used is ``from xrdsst.resources.texts import texts`` and according to  https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3-3 there is no need for _empty_ ``__init.py__`` since 3.3.  Cement framework in use is tested on ``Python 3.5``. Do we want to go in the direction where we try to keep Python version compatibility with lower than 3.5 version?

config/base.yaml
```diff
@@ -12,6 +12,7 @@ security-server:
     owner_member_class: <MEMBER_CLASS>
     owner_member_code: <MEMBER_CODE>
     security_server_code: <SERVER_CODE>
+    software_token_id: 0 # default software token ID is zero
```
bertvi: To conform with the proposed and discussed configuration style, please substitute with <SOFTWARE_TOKEN_ID>

tests/unit/test_token.py
```diff
@@ -0,0 +1,27 @@
+import unittest
+import urllib3
+from xrdsst.controllers.token import TokenController
+
+class TestInit(unittest.TestCase):
```
bertvi: TestInit is the test class created for testing the InitController, please use another name here, e.g. TestToken

tests/unit/test_token.py
```diff
@@ -0,0 +1,27 @@
...
+              'url': 'https://192.168.1.1:8999/api/v1',
+              'api_key': 'X-Road-apikey token=api-key',
+              'configuration_anchor': '/tmp/configuration-anchor.xml',
+              'owner_member_class': 'VOG',
+              'owner_member_code': '4321',
+              'security_server_code': 'SS3',
+              'software_token_id': '0',
+              'software_token_pin': '1122'}]}
+
+    def test_token_list(self):
```
bertvi, line 84: It seems you test here only the negative/failing path, please also test the happy path

tests/unit/test_token.py
```diff
@@ -0,0 +1,27 @@
...
+        token_controller = TokenController()
+        token_controller.load_config = (lambda: self.ss_config)
+        self.assertRaises(urllib3.exceptions.MaxRetryError, lambda: token_controller.list())
+
+    def test_token_login(self):
```
bertvi: It seems you test here only the negative/failing path, please also test the happy path

xrdsst/controllers/init.py
```diff
@@ -62,33 +43,6 @@ def initialize_server(self, config_file):
             else:
                 self.init_security_server(configuration, security_server)
 
-    @staticmethod
```
bertvi, line 65: Due to the refacto of InitServerController, where you moved some base functionality to BaseController, there also now is a need for unit tests for BaseController. Fortunately the tests already exist, please move the unit tests "test_load_config", "test_load_config_exception", "test_init_logging", "test_init_logging_exception" and "test_initialize_basic_conf_values" into the new TestBase class, where you put the unit tests for testing BaseController.

bertvi (changes requested):
Please update project version in "core/version.py" and update the project documentation related to the new token functionality in "docs/xroad_security_server_user_guide.md" and then please drop a line or two in the CHANGELOG.md file about what new has been added/changed in the latest version

config/base.yaml
unserializable: AFAIU it is never different from zero for default software token, so rather would keep the default value, changing it as follows:

```
software_token_id: 0 # <SOFT_TOKEN_ID> -- default software token ID is zero
```
I also thought it might be better to rename it to ``default_software_token_id``, but unprefixed ``software_token_pin`` was already in place, so I decided to withhold that until we have multi-token operations in place, to see how the naming will fit then.

xrdsst/controllers/init.py
unserializable: Commit e16e72764934eb590cc8f588dc7ef5ecd2efd9f6

tests/unit/test_token.py
unserializable: Commit 6a26b61c2dbe2e03c4e001f924f4123a58ff8c62

bertvi (approved):

PR #11, opened 2020-12-08
Sstoolkit 20 21

unserializable:
* implemented first as separate operations, but that was a headache due to interdependent operations and API defs expecting far too much data from plain keys due to OpenAPI definitions
* after sstoolkit-30 discussions of logic application steps converted to use X-Road API single call that creates key on token and its corresponding CSR (which allows immediate SIGN / AUTH differentiation)
* key created on token configured with 'software_token_id' in ``base.yaml`` under ``security-server`` section
* token operation name -- ``init-keys``
* config file additions (unfortunately information not otherwise available):
```yaml
    owner_dn_country: <OWNER_DISTINGUISHED_NAME_COUNTRY>
    owner_dn_org: <OWNER_DISTINGUISHED_NAME_ORGANIZATION>
```
* certificate fields filled according to some conventions of test data, for some fields, e.g. serial number, there might be improvements possible
* conventional key labels used: prefixed with security server name, suffixed with 'default-PURPOSE-key'
* if conventional keys (by label) are found, they will not be dropped or recreated, performs no-op

### DIFF
--- a/tests/unit/test_token.py
+++ b/tests/unit/test_token.py
@@ -11,7 +11,8 @@ from definitions import ROOT_DIR
 from xrdsst.main import XRDSSTTest
 from xrdsst.models import Token, TokenStatus, TokenType, Key, KeyUsageType, TokenCertificate, \
     CertificateOcspStatus, CertificateStatus, CertificateDetails, KeyUsage, CertificateAuthority, \
-    CertificateAuthorityOcspResponse, SecurityServer, KeyWithCertificateSigningRequestId, TokenCertificateSigningRequest
+    CertificateAuthorityOcspResponse, SecurityServer, KeyWithCertificateSigningRequestId, \
+    TokenCertificateSigningRequest
 
 from xrdsst.controllers.token import TokenController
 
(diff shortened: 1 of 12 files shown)


tests/unit/test_token.py
```diff
@@ -104,6 +151,22 @@ def test_token_login(self):
             token_controller.load_config = (lambda: self.ss_config)
             token_controller.login()
 
+    def test_token_init_keys(self):
```
bertvi, line 156: Here you only test the happy path, please also add tests for failures

tests/unit/test_token.py
```diff
@@ -82,6 +127,8 @@ class TestToken(unittest.TestCase):
               'url': 'https://non.existing.url.blah:8999/api/v1',
               'api_key': 'X-Road-apikey token=api-key',
               'configuration_anchor': configuration_anchor,
+              'owner_dn_country': 'FI',
```
bertvi, line 132: Please also add the new fields to TestBaseController configuration values

bertvi (changes requested):

tests/unit/test_token.py
unserializable: There was one, added another for where no CA-s are available at all.

PR #13, opened 2020-12-15
SSTOOLKIT-18

bertvi:


config/base.yaml
```diff
@@ -1,31 +1,32 @@
-### Security Server Toolkit Configuration Settings
-
+api-key:
+- key: /path/to/ssh_private_key
```
unserializable: - Documentation mentions  ``XROAD_SYSTEM_ADMINISTRATOR``, IIRC ``XROAD_SECURITY_OFFICER`` is also required (token ops). Put the defaults into the configuration here, not just placeholders.
- Is there any scenario, where default URL would NOT give desired result, that this has to be configured at all?
- ``key`` preceded by ``api-key`` does not seem intuitive. Something along the lines of 'host-access-key' maybe?

reset_security_server.sh
```diff
@@ -59,19 +57,17 @@ run_ansible_script() {
...
+add_public_key() {
     names=$(echo "$1" | tr "," "\n")
     for name in $names
     do
-        printf "\nCreating api-key for LXD container %s\n" "$name"
-        FILE=api-key-$name.txt
-        lxc exec "$name" -- curl -X POST -u xrd:secret --retry 30 --retry-connrefused --silent "$2" --data "$3" --header "$4" -k -o "$FILE"
-        lxc exec "$name" -- cat api-key-"$name".txt | jq '.key'
+        printf "\nAdding public key to LXD container %s\n" "$name"
+        lxc exec "$name" -- bash -c "echo \"$(cat "$2")\" > .ssh/authorized_keys"
```
unserializable: Here it failed for me, with message:

```
Adding public key to LXD container ss3
bash: .ssh/authorized_keys: No such file or directory
```

Probably because the ``.ssh`` folder is not present in container yet.

Command line was:
```./reset_security_server.sh  -n ss3 -h ../../X-Road/ansible/hosts/lxd_hosts.txt  -a ../../X-Road/ansible -k ~/.ssh/id_ed25519.pub```

and it would seem to me that the content of the public key should be taken from the host and then passed through to  a container.

Comment (bertvi) [reply reset_security_server.sh]: Works for me with the following command:

./reset_security_server.sh -n ss3,ss4 -h ../X-Road/ansible/hosts/lxd_hosts.txt -a ../X-Road/ansible -k ../../work/gofore_public_key

The process:

1. Deletes the given lxd containers if present
2. Installs security servers to the containers using Ansible script and hosts file
3. Copies your public key to ".ssh/authorized_keys" on the containers

File: config/base.yaml (hunk 1 above)
Comment (bertvi) [reply config/base.yaml]: Probably the roles can be as default in the configuration file as these are really the same with each run. The url is probably the same aswell and I could change the name of the "key" parameter here, could be a bit misleading.

File: reset_security_server.sh (hunk 2 above)
Comment (bertvi) [reply reset_security_server.sh]: Probably adding "mkdir .ssh" would solve the problem, I will try adding that there.

File: config/base.yaml (hunk 1 above)
Comment (bertvi) [reply config/base.yaml]: It has been agreed, that placeholders will be used for values in the configuration files. The documentation of how the configuration and running of the toolkit will happen will be further addressed in SSTOOLKIT-6, when we create the User Guide for the Toolkit. I would not place any default values there also in case something will be changed for the SS API, for example some other role has to be used, then we will just need to update the user guide. For the same reason, I would still keep the "url" there, as some parts of that may change and if we would hard-code it to somewhere else, it would not be so simple to change (when for example the url will be changed to https://localhost:5000... or https://localhost:4000/api/v2...). Maybe we should also document about this url in the User Guide, that this is for just creating the api keys and there is no need to change it by users.

File: reset_security_server.sh (hunk 2 above)
Comment (bertvi) [reply reset_security_server.sh]: Added in the code the part that creates the folder when it does not exist

File: config/base.yaml (hunk 1 above)
Comment (unserializable) [reply config/base.yaml]: Ok, I am fine with defaults on hold now, until SSTOOLKIT-6 -- however these values are very unlikely to change in most configurations within _years_ to come -- note that removal of the configuration was not suggested, just simplifying the configuration hoops for cross-referencing documentation by providing default values.

I myself have configuration copy at hand and it does not affect me very much, lets' see what feedback the first users will give. Note that this new configuration is such that placeholders do not allow straightforward replacement anymore, as ``SECURITY_SERVER_ROLE_NAME`` takes multiple values and is listed twice.

**Comment by unserializable** [comment]
So the case with CentOS was that container was lacking ssh server entirely, script finished with following message:
```
Ansible script finished installing security servers 

Adding public key to LXD container ss3
```
although API key was not created. That situation should clearly indicate failure.

File: xrdsst/controllers/base.py [hunk 3]
```diff
@@ -29,36 +62,64 @@ def render(self, render_data):
     def is_output_tabulated(self):
         return self.app.output.Meta.label == 'tabulate'
 
+    def get_api_key(self, conf, security_server):
+        if conf is None:
```
unserializable: More Pythonic would be ``config = conf if conf else self.config``

xrdsst/controllers/base.py
```diff
@@ -19,6 +20,38 @@ class Meta:
...
         ]
 
+    config_file = "config/base.yaml"
+    config = None
+    api_key_default = "X-Road-apikey token=<API_KEY>"
+    api_key_id = {}
+
+    def create_api_key(self, roles_list, config, security_server):
+        self.log_info('Creating API key for security server: ' + security_server['name'])
+        roles = '[\\"'
```
unserializable: This ``json`` acrobatics can probably made much simpler by creating Python list of strings and then converting that to json with ``json.dumps``

```
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> roles = [ "role1", "role2" ]
>>> roles.append("roleN")
>>> json.dumps(roles)
'["role1", "role2", "roleN"]'
```

```

File: xrdsst/controllers/base.py [hunk 5]
```diff
@@ -19,6 +20,38 @@ class Meta:
...
+        count = 1
+        for role in roles_list:
+            if count < len(roles_list):
+                roles += role + '\\",\\"'
+            else:
+                roles += role + '\\"]'
+            count += 1
+        curl_cmd = "curl -X POST -u xrd:secret --silent " + config["api-key"][0]["url"] + " --data \'" + \
```
unserializable: Hardcoding user/password here is not future or real-life compatible.

xrdsst/controllers/token.py
```diff
@@ -54,11 +53,12 @@ def init_keys(self):
...
-    def token_list(self, configuration):
-        self.init_logging(configuration)
-        for security_server in configuration["security-server"]:
-            configuration = self.initialize_basic_config_values(security_server)
+    def token_list(self, config):
+        self.init_logging(config)
+        for security_server in config["security-server"]:
+            configuration = self.initialize_basic_config_values(security_server, config)
             self.remote_token_list(configuration)
+            self.revoke_api_key(security_server, config)
```
unserializable: Please refactor scattered ``revoke_api_key`` calls to aspect. One possible approach would be to register created API keys and then perform revocation on app shutdown. Cement hooks (https://docs.builtoncement.com/core-foundation/hooks) would be one way to realized the aspect, most fitting probably being ``pre_close`` or ``post_close``.

unserializable:
Ok, functionality seems to work nicely. When no api-key is added to configuration file the creation and revocation works (did token login, timestamp service selection, key/csr creation without problems). 

Only small problem seems to be that when no api key is added to configuration file and no-op command is given, it still tries to revoke api-key. Probably it should not do anything in that case and just skip revocation, should be able to determine it by e.g. the API key format.

Situation: config file like that (all previously mentioned ops work nicely), no-op gives error on revocation:
```
api-key:
- credentials: xrd:secret
  key: /HOME/USER/REMOVED
  roles:
  - XROAD_SYSTEM_ADMINISTRATOR
  - XROAD_SECURITY_OFFICER
  url: https://localhost:4000/api/v1/api-keys
logging:
- file: /home/uns/dev/niis/xrdsst-logs/xrdsst-rvw.log
  level: INFO
security-server:
  configuration_anchor: /HOME/REMOVED.xml
  name: ss4
  owner_dn_country: FI
  owner_dn_org: UNSERE
  owner_member_class: GOV
  owner_member_code: 9876
  security_server_code: UNS-SS4
  software_token_id: 0
  software_token_pin: 2345
  url: https://ss4:4000/api/v1
```

```
$ xrdsst
usage: xrdsst [-h] [-d] [-q] [-o {json,tabulate}] [-v] {init,token,timestamp} ...

A toolkit for configuring security server

optional arguments:
  -h, --help            show this help message and exit
  -d, --debug           full application debug mode
  -q, --quiet           suppress all console output
  -o {json,tabulate}    output handler
  -v, --version         show program's version number and exit

sub-commands:
  {init,token,timestamp}
    init                init controller
    token               token controller
    timestamp           timestamp controller
Revoking API key for security server ss4
  File "/home/USER/X-Road-Security-Server-toolkit/env/bin/xrdsst", line 11, in <module>
  File "/home/USER/X-Road-Security-Server-toolkit/xrdsst/main.py", line 87, in main
    app.exit_code = 0
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 1683, in __exit__
    self.close()
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 987, in close
    for res in self.hook.run('pre_close', self):
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/hook.py", line 169, in run
    res = hook[2](*args, **kwargs)
  File "/home/USER/X-Road-Security-Server-toolkit/xrdsst/main.py", line 31, in revert_api_key
    config["api-key"][0]["url"] + "/" + str(api_key_id[security_server['name']]) + " -k"
KeyError: 'ss4'
```
EDIT:

I also experimented with creating API-KEY from console and putting its place into config, like this:
```
...
security-server:
- api_key: X-Road-apikey token=1f2798bf-5c3e-4c38-a562-a4f38656d44e
...
```

it appears it is not taking precedence, the no-op ``xrdsst`` behaves with same minor glitch as before, but the glitch also carries over to operations e.g. ``$ xrdsst token list``

```
$ xrdsst token list
Configuration file "/home/USER/xrdsst-rvw.log" not found: 

API key for security server: ss4 has already been created

|   ID | NAME        | STATUS   | LOGIN STATUS   |
|------+-------------+----------+----------------|
|    0 | softToken-0 | OK       | True           |

Revoking API key for security server ss4
  File "/home/USER/X-Road-Security-Server-toolkit/env/bin/xrdsst", line 11, in <module>
  File "/home/USER/X-Road-Security-Server-toolkit/xrdsst/main.py", line 87, in main
    app.exit_code = 0
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 1683, in __exit__
    self.close()
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 987, in close
    for res in self.hook.run('pre_close', self):
  File "/home/USER/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/hook.py", line 169, in run
    res = hook[2](*args, **kwargs)
  File "/home/USER/X-Road-Security-Server-toolkit/xrdsst/main.py", line 31, in revert_api_key
    config["api-key"][0]["url"] + "/" + str(api_key_id[security_server['name']]) + " -k"
KeyError: 'ss4'
```

PR #15, opened 2020-12-21
Create e2e tests

bertvi:


config/base.yaml
```diff
@@ -1,4 +1,4 @@
-api-key:
+api_key:
```
unserializable, line 1: Can you tell me what has changed here, this is one of the many lines in this PR with no visible changes, but diff highlighted. Could it be that it is some nasty CR/LF/CRLF thing only? If so, we should probably configure our git to keep these under control.

run_end_to_end_tests.sh
```diff
@@ -0,0 +1,62 @@
...
+  echo "Usage: run_end_to_end_tests.sh -c config_file -a configuration_anchor -h security_server_host -n security_server_name -k private_key_file"
+}
+
+exit_abnormal() {
+  usage
+  exit 1
+}
+
+update_config() {
+  cmd=".api_key[0].key=\"$5\""
```
unserializable, line 45: This ``cmd`` variable name is very prone to clashing sooner or later, since interpreter is explicitly set as ``bash``, I would recommend to declare this variable local.

```bash
local cmd
cmd=""
cmd="${cmd}|...."
```

config/base.yaml
bertvi: I have updated all the configuration file parameters to use underscores instead of dashes as dashes create problems in json and yaml, got some errors when using YQ for creating a new YAML configuration from template.

run_end_to_end_tests.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+# Usage prepare_end_to_end_tests.sh -c config_file
```
unserializable: The usage sample name ``prepare_end_to_end_tests.sh`` differs from file name `` run_end_to_end_tests.sh``

run_end_to_end_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+#   -c: config file
+#   -a: configuration-anchor file
+#   -h: host name or IP address of the security server
+#   -n: security server name
+#   -k: private ssh key file
+#   -u: credentials
+#   -o: output file
+#
+#
+# Usage example: prepare_end_to_end_tests.sh -c tests/resources/test-config.yaml
```
unserializable: The usage sample name ``prepare_end_to_end_tests.sh`` differs from file name `` run_end_to_end_tests.sh``

run_end_to_end_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+#                                             -a /etc/xroad/configuration_anchor.xml
+#                                             -h ss
+#                                             -n ss
+#                                             -k /home/user/id_rsa
+#                                             -u xrd:secret
+#                                             -o tests/resources/test-config.yaml
+
+
+usage() {
+  echo "Usage: prepare_end_to_end_tests.sh -c config_file
```
unserializable: The usage sample name ``prepare_end_to_end_tests.sh`` differs from file name `` run_end_to_end_tests.sh``

run_end_to_end_tests.sh
```diff
@@ -0,0 +1,95 @@
...
+}
+
+update_config() {
+  cmd=".api_key[0].key=\"$5\""
+  cmd="${cmd}|.security_server[0].configuration_anchor=\"$2\""
+  cmd="${cmd}|.security_server[0].name=\"$4\""
+  cmd="${cmd}|.security_server[0].security_server_code=\"$4\""
+  cmd="${cmd}|.security_server[0].url=\"https://$3:4000/api/v1\""
+  cmd="${cmd}|.api_key[0].credentials=\"$6\""
+  yq -y "$cmd" "$1" > "$7"
```
unserializable, line 51: This probably could use some dependency remark in the script, my Ubuntu does not have ``yq`` installed & it is not quite obvious which is the best way to install it.

unserializable: Hmm, also apparantly ``yq`` v3 and v4 are rather incompatible, which one is meant for using here (v4 is still in beta?)?

unserializable: Just a thought, /if/ we are not doing anything particularly complex to the yaml here, might be better off skipping ``yq`` and going with simple substitution in bash 'heredoc,  see https://linuxize.com/post/bash-heredoc/ & https://tldp.org/LDP/abs/html/here-docs.html

unserializable: So I found that ``yq`` is available as snap and installed it, but script failed as ``yq`` does not recognize ``-y`` parameter:

```
Error: unknown shorthand flag: 'y' in -y
```
version report

```bash
$ yq --version
yq version 3.4.1
``

Comment (unserializable) [reply run_end_to_end_tests.sh:51]: After the commit fb4352dcd2b254fc875cd1a1608a097c9adb4140 and after i  **uninstall** ``yq`` and run the script in virtualenv there are no error messages but the output config file is not created either.

File: run_end_to_end_tests.sh [hunk 7]
```diff
@@ -0,0 +1,95 @@
...
+#   -o: output file
+#
+#
+# Usage example: prepare_end_to_end_tests.sh -c tests/resources/test-config.yaml
+#                                             -a /etc/xroad/configuration_anchor.xml
+#                                             -h ss
+#                                             -n ss
+#                                             -k /home/user/id_rsa
+#                                             -u xrd:secret
+#                                             -o tests/resources/test-config.yaml
```
unserializable: Is it intentional that input config file (``-c``) and output file (``-o``) are the same here?

run_end_to_end_tests.sh
unserializable: This one here refers to ``prepare_end_to_end_tests.sh`` still

unserializable: And also it lists the ``-o`` output parameter

PR #16, opened 2020-12-22
Sstoolkit m32 33

unserializable:
* Implements import for certificates
* Implements registration request for authentication certificate


config/base.yaml
```diff
@@ -10,6 +10,9 @@ logging:
   level: <LOG_LEVEL>
 security_server:
 - api_key: X-Road-apikey token=<API_KEY>
+  certificates:
```
bertvi, line 13: you have placed the "certificates" section only for the first security server, what about the other one?

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -86,6 +98,9 @@ The ``security-server`` section is for configuring security server parameters
 * <SOFT_TOKEN_ID> default software token ID, normally 0 (zero).
 * <SOFT_TOKEN_PIN> should be substituted with a desired numeric pin code
 * <SECURITY_SERVER_FQDN_OR_IP> should be substituted with the IP address or host name of the installed security server, e.g. ss1
+* ``/path/to/signcert`` and ``/path/to/authcert`` should be given as paths referring to certificate locations,
```
bertvi, line 101: In order for the documentation style to be consistent with line 101 in the User Guide, please also update lines 86, 88 and 91 accordingly

bertvi (changes requested):
1. "tests/unit/test_base_controller": here the configuration settings differ from the ones in "base.yaml", please update these aswell

2. Please also update the changelog

3. "controllers/cert.py": 
 * unused import on line 1
 * the comments on lines 13-24 seem inappropriate here, especially the "acceptance criteria". If the functionality needs some extra explanation, please place it in the User Guide
 
4. Please enable "PEP 8 coding style violation" and "PEP 8 naming convention violation" under "Settings->Editor->Inspections->Python" in PyCharm - for me it shows several violations in "controllers/cert.py", "controllers/token.py", "tests/unit/test_cert.py", "tests/unit/test_token.py" and "tests/unit/test_xrdsst.py", probably most of these could be fixed.
 
5. When running "make test", all tests seem to pass, but when running "test_cert.py" individually from the file, all tests except "test_cert_register_nonresolving_url" fail, tests should pass either way - individually from test class or by using Makefile target.

config/base.yaml
unserializable: 1. Removed irrelevant config from ``test_base_controller``
2. Added 
3. Removed 
4. These are enabled, couple of them indeed i had not noticed and fixed them. Note that I have ignored some of the more ones and these checks do not completely match the ones used during build process.
5. Cannot reproduce single test run failure, please give command-line failure example.

unserializable (commented):
@bertvi One of your comments was accidentally deleted when I deleted my own in-edit _reply_ to it, it was about cerfiticates section for the second security server in config file -- it is optional as most of these, so saw no point in adding it to second. It makes no sense to have identical configs anyway, maybe we could strip this second one to bare minimum to provide useful example (all the cert defs, services defs, etc are something that users will not have until working with toolkit for a while).

unserializable:
>     1. "tests/unit/test_base_controller": here the configuration settings differ from the ones in "base.yaml", please update these aswell
> 
>     2. Please also update the changelog
> 
>     3. "controllers/cert.py":
> 
> 
>     * unused import on line 1
> 
>     * the comments on lines 13-24 seem inappropriate here, especially the "acceptance criteria". If the functionality needs some extra explanation, please place it in the User Guide
> 
> 
>     1. Please enable "PEP 8 coding style violation" and "PEP 8 naming convention violation" under "Settings->Editor->Inspections->Python" in PyCharm - for me it shows several violations in "controllers/cert.py", "controllers/token.py", "tests/unit/test_cert.py", "tests/unit/test_token.py" and "tests/unit/test_xrdsst.py", probably most of these could be fixed.
> 
>     2. When running "make test", all tests seem to pass, but when running "test_cert.py" individually from the file, all tests except "test_cert_register_nonresolving_url" fail, tests should pass either way - individually from test class or by using Makefile target.

The reply to this seems to be lost, it was roughly:

1. Removed irrelevant configs from ``test_base_controller``
2. Updated
3. Removed
4. These are enabled and what I saw were addressed, they differ from ones in build process somewhat.
5. This does not seem to happen, please provide repro details if problem persists.

bertvi:
> @bertvi One of your comments was accidentally deleted when I deleted my own in-edit _reply_ to it, it was about cerfiticates section for the second security server in config file -- it is optional as most of these, so saw no point in adding it to second. It makes no sense to have identical configs anyway, maybe we could strip this second one to bare minimum to provide useful example (all the cert defs, services defs, etc are something that users will not have until working with toolkit for a while).

If you wish to keep the "certificates" section in just one of the security server sections, then please delete the other security server section. Otherwise it may be confusing. If users want to have multiple, they will add new sections themselves.

unserializable:
> > ```
> > 5. When running "make test", all tests seem to pass, but when running "test_cert.py" individually from the file, all tests except "test_cert_register_nonresolving_url" fail, tests should pass either way - individually from test class or by using Makefile target.
> > ```
>     5. This does not seem to happen, please provide repro details if problem persists.

Ok, so it appears you were running single tests with ``unittest`` instead of ``pytest``, tests use I/O capture fixtures of PyTest and mixing fixtures with unit tests seems to be only supported via ``pytest`` runner that we usually use.  

https://docs.pytest.org/en/stable/unittest.html#mixing-fixtures

> Running your unittest with pytest allows you to use its fixture mechanism with unittest.TestCase style tests.

Single tests run fine with ``pytest``:

```
X-Road-Security-Server-toolkit/tests/unit $ python -m pytest  test_cert.py::TestCert::test_cert_register_multiple_certs_for_auth_key
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.8.5, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir:  /X-Road-Security-Server-toolkit
plugins: pylint-0.18.0, cov-2.10.1
collected 1 item                                                                                                                                                                                                  

test_cert.py Configuration file "/tmp/xrdsst_test_token_log" not found: 

Starting configuration process for security server: ssX
API key for security server: ssX has already been created
Multiple registrable certificates for key labelled 'ssX-default-auth-key'.
.                                                                                                                                                                                              [100%]

================================================================================================ 1 passed in 0.53s ================================================================================================
```

bertvi (approved):

PR #18, opened 2021-01-07
Sstoolkit 35

unserializable:
* Separate subsystem creation with ``xrdsst client add``
* Fix log file not being created
* Bit more verbosity for token login and cert import operations.


xrdsst/controllers/base.py
```diff
@@ -87,15 +87,18 @@ def get_api_key(self, conf, security_server):
...
-            log_file_name = configuration["logging"][0]["file"]
-            if not os.path.isfile(log_file_name):
-                raise FileNotFoundError
-            logging.basicConfig(filename=log_file_name,
-                                level=configuration["logging"][0]["level"],
-                                format='%(name)s - %(levelname)s - %(message)s')
-        except FileNotFoundError as err:
-            print("Configuration file \"" + log_file_name + "\" not found: %s\n" % err)
+            exists = os.path.exists(log_file_name)
+            if exists and os.path.isdir(log_file_name):
```
bertvi, line 94: If non-existing path is given, nothing is reported to the user and nothing gets logged

xrdsst/controllers/base.py
```diff
@@ -145,3 +148,10 @@ def security_server_address(security_server):
         :return: IP/host deduced from security server URL
         """
         return urlparse(security_server['url']).netloc.split(':')[0]  # keep the case, unlike with '.hostname'
+
+    @staticmethod
+    def convert_swagger_enum(type, value):
```
bertvi: Change to "_type" as you have done in cert.py

xrdsst/controllers/cert.py
```diff
@@ -114,9 +114,10 @@ def remote_import_certificates(ss_configuration, security_server):
                     cert_data = cert_file.read()
                     cert_file.close()
                     token_cert_api.import_certificate(body=cert_data)
+                    BaseController.log_info("Imported certificate '" + certfile + "'")
                 except ApiException as err:
                     if err.status == 409 and err.body.count("certificate_already_exists"):
-                        print("Certificate '" + certfile + "' already imported.")
+                        BaseController.log_info("Certificate '" + certfile + "' already imported.")
```
bertvi: certfile and cert_file could cause ambiguity, you could change certfile to cert_filename for example.

xrdsst/controllers/client.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @ex(help="Add client subsystem", arguments=[])
+    def add(self):
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+        self.add_client(self.load_config())
+
+    def add_client(self, configuration):
+        self.init_logging(configuration)
+        for security_server in configuration["security_server"]:
+            BaseController.log_info('Starting configuration process for security server: ' + security_server['name'])
```
bertvi: Exactly same information gets logged as in InitController, please consider changing a bit to be more clear what is configured here.

bertvi (changes requested):

xrdsst/controllers/cert.py
unserializable: -> cert_file_loc

xrdsst/controllers/client.py
unserializable: This is the same everywhere, token, client, cert -- do you have actual suggestion to approach this, before the autoconfiguration command is also implemented?

bertvi: I have not noticed it before. Seems to be the same in init, cert, client and token. In all others except init, please add just one or two words what kind of cofiguration process is being started, for token e.g. "Starting token configuration...", otherwise we just have a lot of similar lines in the logs.

unserializable: Redundancy in autoconfiguration is such that I do not believe any of these messages will survive. However, I amended this ATM.

bertvi (approved):

PR #19, opened 2021-01-11
add s3pypi dependency for uploading packages

bertvi:


unserializable (changes requested):
Build does complain of conflicting dependencies:

```
...
ERROR: s3pypi 0.11.0 has requirement wheel<0.34.0,>=0.33.6, but you'll have wheel 0.34.2 which is incompatible.
...
```

Since there is nothing using the s3pypi here in this project, I would recommend dropping this  change and closing PR.

If the ``s3pypi`` will be for some reason required *inside* the project, the compatibility of dependencies should be ensured and tested & most likely its functionality should be made accessible via ``Makefile`` target.

PR #20, opened 2021-01-14
Sst36 t2

unserializable:
Adds subsystem/client registration functionality.

* API key used must have ``XROAD_REGISTRATION_OFFICER`` role for subsystem adding / registration
  
To run the integration tests, central server needs to have:
 * ORG:GOV:1234 member defined.
 * autoconfiguration must be enabled in center.ini (or local.ini?) 
```
   [center]
   auto-approve-auth-cert-reg-requests=true
   auto-approve-client-reg-requests=true
```
 * management services via security server must be configured, with client access for ``security-server-owners``
 * management services security server must be accessible via IP address
 * management services security server must be accessible from Docker container

### DIFF
--- a/tests/integration/test_xrdsst.py
+++ b/tests/integration/test_xrdsst.py
@@ -26,7 +26,7 @@ def waitfor(boolf, delay, retries):
         count += 1
 
     if count >= retries:
-        raise Exception("Exceeded retry count " + retries + " with delay " + delay)
+        raise Exception("Exceeded retry count " + str(retries) + " with delay " + str(delay) + "s.")
 
 
 # Returns TEST CA signed certificate in PEM format.
--- a/tests/integration/test_xrdsst.py
+++ b/tests/integration/test_xrdsst.py
@@ -242,6 +242,13 @@ class TestXRDSST(unittest.TestCase):
             client_controller.load_config = (lambda: self.config)
             client_controller.add()
 
+    def step_subsystem_register(self):
+        with XRDSSTTest() as app:
+            client_controller = ClientController()
+            client_controller.app = app
+            client_controller.load_config = (lambda: self.config)
+            client_controller.register()
+
     def test_run_configuration(self):
         urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
         self.step_init()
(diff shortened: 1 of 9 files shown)


tests/integration/test_xrdsst.py
```diff
@@ -17,15 +17,27 @@
...
+
+# Waits until boolean function returns True within number of retried delays or raises error
+def waitfor(boolf, delay, retries):
+    count = 0
+    while count < retries and not boolf():
+        time.sleep(delay)
+        count += 1
+
+    if count >= retries:
+        raise Exception("Exceeded retry count " + retries + " with delay " + delay)
```
bertvi: TypeError: can only concatenate str (not "int") to str

xrdsst/controllers/client.py
```diff
@@ -49,6 +62,43 @@ def remote_add_client(ss_configuration, client_conf):
             else:
                 BaseController.log_api_error('ClientsApi->add_client', err)
 
+    @staticmethod
+    def remote_register_client(ss_configuration, security_server_conf, client_conf):
+        client = None
```
bertvi: Unused variable

tests/integration/test_xrdsst.py
bertvi: After setting auto-configuration options in center.ini for my central server running in an LXD container, I get the a timeout and maximum retries of 300 exceeded when running "make test-all"

bertvi (changes requested):

tests/integration/test_xrdsst.py
unserializable: What time limit works for your setup and was central server restarted?

bertvi: > What time limit works for your setup and was central server restarted?

Seems not to work at all. I have restarted my CS instance, set retries to 600. Does it work for you?

bertvi: > What time limit works for your setup and was central server restarted?

If it works for you, I will try to make it work as well, but you can go ahead and merge.

bertvi (approved):

tests/integration/test_xrdsst.py
unserializable: Yes it does, too well even, as one code section that raised exception when retry limit was exceeded had been never reached, so it had a missing typecast there. Let me know if problem persists and let's sort this out.

Merging.

PR #22, opened 2021-01-22
SSTOOLKIT-38 and SSTOOLKIT-39

bertvi:
PR for SSTOOLKIT-38 and SSTOOLKIT-39 combined

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 
-## [0.1.21-alpha.0] - 2021-01-21
+## [0.2.0-alpha.0] - 2021-01-21
 
 - add ``xrdsst client`` subcommand ``add-service-description``
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.1.20-alpha.0] - 2021-01-21
+
+- add ``xrdsst client`` subcommand ``add-service-description``
+
 ## [0.1.19-alpha.0] - 2021-01-12
 
 - add ``xrdsst client`` subcommand ``register``
 
 ## [0.1.18-alpha.0] - 2021-01-07
 
-- add ``xrdsst client`` and subcommand ``add``
+- add ``xrdsst client`` and subcommand ``add-subsystem``
 
 ## [0.1.17-alpha.0] - 2020-12-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ All notable changes to this project will be documented in this file.
 
 ## [0.1.18-alpha.0] - 2021-01-07
 
-- add ``xrdsst client`` and subcommand ``add-subsystem``
+- add ``xrdsst client`` and subcommand ``add``
 
 ## [0.1.17-alpha.0] - 2020-12-31
 
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -5,6 +5,7 @@ api_key:
   - <SECURITY_SERVER_ROLE_NAME>
   - <SECURITY_SERVER_ROLE_NAME>
   - <SECURITY_SERVER_ROLE_NAME>
+  - <SECURITY_SERVER_ROLE_NAME>
   url: https://localhost:4000/api/v1/api-keys
 logging:
 - file: /path/to/xrdsst.log
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -32,3 +32,4 @@ security_server:
       service_descriptions:
         - url: <SERVICE_DESCRIPTION_URL>
           rest_service_code: <REST_SERVICE_CODE>
+          type: <SERVICE_TYPE>
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -28,3 +28,6 @@ security_server:
       member_code: <MEMBER_CODE>
       subsystem_code: <SUBSYSTEM_CODE>
       connection_type: <CONNECTION_TYPE>
+      service_descriptions:
+        - url: <SERVICE_DESCRIPTION_URL>
+          rest_service_code: <REST_SERVICE_CODE>
--- a/config/base.yaml
+++ b/config/base.yaml
@@ -1,13 +1,7 @@
-logging:
-  file: /path/to/xrdsst.log
-  level: <LOG_LEVEL>
 api_key:
 - credentials: xrd:secret
   key: /path/to/ssh_private_key
   roles:
-  - <SECURITY_SERVER_ROLE_NAME>
-  - <SECURITY_SERVER_ROLE_NAME>
-  - <SECURITY_SERVER_ROLE_NAME>
   - <SECURITY_SERVER_ROLE_NAME>
   url: https://localhost:4000/api/v1/api-keys
 security_server:
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -167,6 +167,6 @@ cert register``, final activation with ``xrdsst cert activate``.
 
 ### 3.7 Client management
 Client subsystems are managed with ``xrdsst client`` subcommands, new subsystem client can be added with
-``xrdsst client add-subsystem``, the subsystem parameters should be specified in the configuration ``clients`` section.
+``xrdsst client add``, the subsystem parameters should be specified in the configuration ``clients`` section.
 Further subsystem registration can proceed with ``xrdsst client register``. Adding REST/OPENAPI3/WSDL service descriptions
 is performed with ``xrdsst client add-service-description``.
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -86,7 +86,7 @@ security-server:
       connection_type: <CONNECTION_TYPE>
       service_descriptions:
         - url: <SERVICE_DESCRIPTION_URL>
-          rest_service_code: <SERVICE_CODE>
+          rest_service_code: <REST_SERVICE_CODE>
           type: <SERVICE_TYPE>
 ```
 
@@ -116,7 +116,7 @@ in fact any number of certificates can be imported for the keys labelled ``defau
 * <SUBSYSTEM_CODE> X-Road member/client subsystem code.
 * <CONNECTION_TYPE> Connection protocol selection, from among ``HTTP``, ``HTTPS``, ``HTTPS_NO_AUTH``.
 * <SERVICE_DESCRIPTION_URL> URL for service description
-* <SERVICE_CODE> rest service code, empty for WSDL service
+* <REST_SERVICE_CODE> rest service code, not used for WSDL services
 * <SERVICE_TYPE> type of service, value from ``OPENAPI3``, ``REST``, ``WSDL``.
 
 ## 3 Running the X-Road Security Server Toolkit
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -2,7 +2,7 @@
 
 **Technical Specification**
 
-Version: 1.1.6
+Version: 1.1.7
 Doc. ID: XRDSST-CONF
 
 | Date       | Version     | Description                                                                  | Author             |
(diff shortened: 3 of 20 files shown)


CHANGELOG.md
```diff
@@ -12,7 +17,7 @@ All notable changes to this project will be documented in this file.
 
 ## [0.1.18-alpha.0] - 2021-01-07
 
-- add ``xrdsst client`` and subcommand ``add``
+- add ``xrdsst client`` and subcommand ``add-subsystem``
```
unserializable: Long this command name is now .... What is the motivation behind the command name change, as far as I remember we decided that client is the correct term to use and plain 'add' covers the cases?

xrdsst/controllers/client.py
```diff
@@ -52,48 +66,76 @@ def remote_add_client(ss_configuration, client_conf):
...
         except ApiException as err:
             if err.status == 409:
                 BaseController.log_info("Client for '" + partial_client_id(client_conf) + "' already exists.")
             else:
                 BaseController.log_api_error('ClientsApi->add_client', err)
 
-    @staticmethod
-    def remote_register_client(ss_configuration, security_server_conf, client_conf):
+    def remote_add_service_description(self, ss_configuration, security_server_conf, client_conf, service_description_conf):
+        code = service_description_conf['rest_service_code'] if service_description_conf['rest_service_code'] else None
```
unserializable, line 76: This does not work as intended for avoiding KeyError, use dictionary .get('key') :)

xrdsst/controllers/client.py
```diff
@@ -52,48 +66,76 @@ def remote_add_client(ss_configuration, client_conf):
...
-            )
+            client = self.find_client(clients_api, security_server_conf, client_conf)
+            if client:
+                try:
+                    response = clients_api.add_client_service_description(client.id, body=description_add)
+                    if response:
+                        BaseController.log_info("Added client subsystem " + partial_client_id(client_conf) + " service description" +
+                                                " (got full id " + response.id + ")")
+                except ApiException as err:
+                    if err.status == 409:
```
unserializable, line 91: This information seems not quite sufficient when several service descriptions are used, it will then give quite non-informative messages, please see if it is possible to amend this so that it gives useful feedback.  

```
Service description for 'GOV:1234:OMNIBUS' already exists.
Service description for 'GOV:1234:OMNIBUS' already exists.
```

xrdsst/controllers/timestamp.py
```diff
@@ -78,10 +76,11 @@ def remote_ts_list(self, apicall):
         try:
             ts_list_response = apicall()
             self.render_timestamping_services(ts_list_response)
-        except ApiException as e:
-            print("Exception when listing timestamping services: %s\n", e)
+        except ApiException as exc:
+            self.log_api_error('Exception when listing timestamping services:', exc)
```
unserializable: We once came to agreement to not log anything from this, as it is read-only operation that is only explicitly invoked from command-line and there is still the comment about that at  ``timestamp_service_list``. I do not see a reason to change it, why was this done?

xrdsst/controllers/timestamp.py
```diff
@@ -92,16 +91,15 @@ def remote_timestamp_service_list(self, ss_configuration):
...
-    def remote_get_configured(ss_configuration):
+    def remote_get_configured(self, ss_configuration):
         try:
             system_api = SystemApi(ApiClient(ss_configuration))
             ts_list_response = system_api.get_configured_timestamping_services()
             return ts_list_response
-        except ApiException as e:
-            print("Exception when listing timestamping services: %s\n", e)
+        except ApiException as exc:
+            self.log_api_error('Exception when listing timestamping services:', exc)
```
unserializable: This seems not to have any use for logging as well, as the method is used in explicit read-only ops / tests,

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -111,6 +115,9 @@ in fact any number of certificates can be imported for the keys labelled ``defau
 (but not all of them can be in use / registered)
 * <SUBSYSTEM_CODE> X-Road member/client subsystem code.
 * <CONNECTION_TYPE> Connection protocol selection, from among ``HTTP``, ``HTTPS``, ``HTTPS_NO_AUTH``.
+* <SERVICE_DESCRIPTION_URL> URL for service description
+* <SERVICE_CODE> rest service code, empty for WSDL service
```
unserializable: It seems this is better off completely being not used for WSDL service.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file.
 
+
+## [0.2.0-alpha.0] - 2021-01-21
+
+- add ``xrdsst client`` subcommand ``add-service-description``
```
unserializable, line 8: For me, the command name that seemed to click better was ``describe-services`` -- how do you feel about that?

config/base.yaml
```diff
@@ -8,6 +8,7 @@ api_key:
   - <SECURITY_SERVER_ROLE_NAME>
   - <SECURITY_SERVER_ROLE_NAME>
   - <SECURITY_SERVER_ROLE_NAME>
+  - <SECURITY_SERVER_ROLE_NAME>
```
unserializable: This section is getting sillier as the roles are added, how about just adding them all here and be done with them or avoiding the repetition a'la:

```
-  <SECURITY_SERVER_ROLE_NAME>
-  # ...  <SECURITY_SERVER_ROLE_NAME> ...
```

requirements.txt
```diff
@@ -18,3 +17,7 @@ gitpython~=3.1.11
 docker~=4.1.0
 yq~=2.11.1
 jq~=1.1.1
+s3pypi~=0.11.0
```
unserializable: Is this return of the s3pypi dependency intentional, we once skipped it in https://github.com/nordic-institute/X-Road-Security-Server-toolkit/pull/19

Still the resolver cannot resolve this dependencies with known good compatibility versions:
 ```
  ERROR: s3pypi 0.11.0 has requirement wheel<0.34.0,>=0.33.6, but you'll have wheel 0.34.2 which is incompatible.
  ```
  
As we still have no code that does anything with s3 it seems we should not include that dependency.

setup.cfg
```diff
@@ -9,7 +9,7 @@ serialize =
 
 [MASTER]
 ignore = rest,models,resources,api,api_client,configuration
-disable = missing-module-docstring,missing-class-docstring,missing-function-docstring,misplaced-comparison-constant,too-few-public-methods,unnecessary-lambda,fixme
+disable = missing-module-docstring,missing-class-docstring,missing-function-docstring,misplaced-comparison-constant,too-few-public-methods,unnecessary-lambda,fixme,E0401
```
unserializable: Please add these ignores in human readable form.

tests/end_to_end/tests.py
```diff
@@ -80,9 +88,122 @@ def step_token_init_keys(self):
...
+            cert_files.append(cert_file)
+            with open(cert_file, "w") as out_cert:
+                out_cert.write(cert)
+
+        return cert_files
+
+    def apply_cert_config(self, signed_certs):
+        self.config['security_server'][0]['certificates'] = signed_certs
+
+    def revoke_api_key(self, security_server):
```
unserializable: I really do not like this, the presence of this function and the many places where it is called from seems rather detrimental to readability here. Could it possibly better to just:
  a) register API key for tests and use that
  b) use the ``tearDown()`` method?
  c) as a bonus for more testing, could the application revocation method be used / tested here

xrdsst/controllers/client.py
```diff
@@ -31,6 +36,15 @@ def add_client(self, configuration):
             for client in security_server["clients"]:
                 self.remote_add_client(ss_configuration, client)
 
+    def add_service_description(self, configuration):
+        self.init_logging(configuration)
+        for security_server in configuration["security_server"]:
+            BaseController.log_info('Starting service description add process for security server: ' + security_server['name'])
+            ss_configuration = self.initialize_basic_config_values(security_server, configuration)
+            for client in security_server["clients"]:
+                for service_description in client["service_descriptions"]:
```
unserializable: I think this should be conditional, there might well be no service descriptions defined yet in which case this fails...

xrdsst/main.py
```diff
@@ -98,10 +98,10 @@ def revoke_api_key(app):
...
                     log_info('Revoking API key for security server ' + security_server['name'])
                     curl_cmd = "curl -X DELETE -u " + config["api_key"][0]["credentials"] + " --silent " + \
-                        config["api_key"][0]["url"] + "/" + str(api_key_id[security_server['name']]) + " -k"
+                               config["api_key"][0]["url"] + "/" + str(api_key_id[security_server['name']]) + " -k"
                     cmd = "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i \"" + \
-                        config["api_key"][0]["key"] + "\" root@" + security_server["name"] + " \"" + curl_cmd + "\""
-                    process = subprocess.run(cmd, shell=True, check=False, capture_output=True)
+                          config["api_key"][0]["key"] + "\" root@" + security_server["name"] + " \"" + curl_cmd + "\""
+                    subprocess.run(cmd, shell=True, check=False, capture_output=True)
                     log_info('API key for security server ' + security_server['name'] + ' revoked successfully')
```
unserializable: Should this not check for run result before claiming successful revocation?

unserializable (changes requested):
Mixture of comments here, their priorities are also mixed but hopefully determinable from comments themselves.

CHANGELOG.md
bertvi: I do not agree

xrdsst/controllers/client.py
bertvi: 1. What is the point of calling add_service_description() with none defined in the configuration file? 2. This is a for loop, which runs when there are some values present.

unserializable: But it fails here with KeyError then, something that later is tried for amending at ``remote_add_service_description``, but without effect due to the call site already requiring element presenece in YAML. If anywhere, this is the place for conditional here.

xrdsst/controllers/base.py
```diff
@@ -154,7 +154,7 @@ def initialize_basic_config_values(self, security_server, config=None):
 
     @staticmethod
     def log_api_error(api_method, exception):
-        logging.error("Exception calling " + api_method + ": " + str(exception))
+        logging.error("Exception calling %s: %s", api_method, str(exception))
```
unserializable: This is something that eluded my notice, since there are quite many formatting changes in this PR. It currently does no harm. Operator ``%`` operator is essentially dead, ``printf`` style format specifiers not. However:

* https://docs.python.org/3/howto/logging.html#logging-variable-data


> ...
> the old, %-style of string formatting. This is for backwards compatibility: the logging package pre-dates newer formatting options such as str.format() and string.Template. These newer formatting options are supported, but exploring them is outside the scope of this tutorial: ...

seems to suggest it is better to either leave it as is or use the Python 3 formatting options.

xrdsst/controllers/client.py
unserializable: Hmm, I wrote previous comment while multitasking, so I am not certain I made the correct code reference. However, simply put, this fails whenever client does not have ``service_descriptions``:

## Config section
```
  clients:
    - member_class: GOV
      member_code: 1234
      subsystem_code: OMNIBUS
      connection_type: HTTP
```

## Result

```bash
$ xrdsst -c config/ss5.yaml client add-service-description 
Starting service description add process for security server: ss5
API key for security server: ss5 has already been created
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/env/bin/xrdsst", line 11, in <module>
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/main.py", line 154, in main
    app.run()
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 916, in run
    return_val = self.controller._dispatch()
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/ext/ext_argparse.py", line 808, in _dispatch
    return func()
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/controllers/client.py", line 25, in add_description
    self.add_service_description(self.load_config())
  File "/home/user/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/controllers/client.py", line 45, in add_service_description
    for service_description in client["service_descriptions"]:
KeyError: 'service_descriptions'
```

xrdsst/controllers/client.py
```diff
@@ -19,6 +19,11 @@ class Meta:
     def add(self):
         self.add_client(self.load_config())
 
+    @ex(label='add-service-description', help="Add client service description", arguments=[])
+    def add_description(self):
+        urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
```
unserializable: This ``urllib3`` insecure request warning can be removed as PR base is up-to-date with PR21 and this is only set once in application -- (for release it seems that one call will be removed too).

PR #23, opened 2021-01-27
SSTOOLKIT-40

bertvi:


xrdsst/controllers/client.py
```diff
@@ -52,6 +56,16 @@ def register_client(self, configuration):
             for client in security_server["clients"]:
                 self.remote_register_client(ss_configuration, security_server, client)
 
+    def enable_service_description(self, configuration):
+        self.init_logging(configuration)
+        for security_server in configuration["security_server"]:
+            BaseController.log_info('Starting service description add process for security server: ' + security_server['name'])
```
unserializable: Wrong operation is logged here.

xrdsst/controllers/client.py
```diff
@@ -52,6 +56,16 @@ def register_client(self, configuration):
...
+            ss_configuration = self.initialize_basic_config_values(security_server, configuration)
+            for client in security_server["clients"]:
```
unserializable: Please make it so that presence of both "clients" and "service_descriptions" are optional in configuration file, when running this command.

tests/integration/test_xrdsst.py
```diff
@@ -248,6 +248,13 @@ def step_subsystem_add_service_description(self):
             client_controller.load_config = (lambda: self.config)
             client_controller.add_description()
 
+    def step_subsystem_enable_service_description(self):
+        with XRDSSTTest() as app:
+            client_controller = ClientController()
+            client_controller.app = app
+            client_controller.load_config = (lambda: self.config)
+            client_controller.enable_description()
```
unserializable: This operation seems to have no later dependent operations at all? If so, please verify that this was indeed successful by querying the security server.

tests/unit/test_client.py
```diff
@@ -233,3 +237,68 @@ def getheaders(self): return None
...
+                                return_value=[ClientTestData.add_description_response]):
+                    with mock.patch('xrdsst.api.service_descriptions_api.ServiceDescriptionsApi.enable_service_description',
+                                    return_value=None):
+                        client_controller = ClientController()
+                        client_controller.app = app
+                        client_controller.load_config = (lambda: self.ss_config)
+                        client_controller.enable_description()
+
+                        out, err = self.capsys.readouterr()
+                        assert out.count("service description") > 0
```
unserializable: Make this assert more specific please, as it stands it also triggers on e.g. 'service description failed'.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.1-alpha.0] - 2021-01-27
+
+- add ``xrdsst client`` subcommand ``enable-service-description``
```
unserializable: The service operations continue coming up and is seems better to do `service` subcommand entirely where to move service and service description specific things. Having it as client subcommand is not that expressive either, as application logic does not have single security server / client context when running service commands, but everything is batched.

For the existing operations then we can also have simple verbs without remembering the placement of dashes and plural and singular form confusions:

```
xrdsst service describe
xrdsst service enable-description
```

Moreover it seems upcoming service rights / parameters tuning and API `/services` calls probably can fit into this same subcommand, but this is not certain, so please give some thought to whether suggestion is good approach.

unserializable (changes requested):

PR #24, opened 2021-02-03
Sstoolkit 49p2

unserializable:
This PR has the following:
  * core/api_util.py
    - status classes for representing server configuration status and status of different server operations 
    - API status query methods that can be used per-operation when dependent operations have verified to have been completed, or unconditionally in succession, in which case they must be grouped according to the server initialization status: timestamping, token, token key, token key csr, token key cert operations API call DO return 500 ``{"status":500,"error":{"code":"core.MalformedServerConf","metadata":["Server conf is not initialized!"]}`` if the initialization has not been completed.
    - status classes should are mostly sufficient to correctly traverse the dependent operation graph depending on the server-side configuration, and accordingly, to present end-user with the context (in much shorter form than this shown)
  * controllers/status.py
    - single console operation 'status' that returns the /core/ configuration status -- this does not include clients or services, the output is already quite long and these things probably do deserve their own listing methods that can both give better overview and be more detailed
    - the status is expected to work in all conditions where security server is reachable at all and API key access has been provider
    - information output in json is bit more detailed than for console, to be more useful for possible scripting (e.g. hashes of toolkit-generated certs, if available)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.2-alpha.0] - 2021-02-03
+- Add examination of configured server statuses with ``xrdsst status``
+
 ## [0.2.1-alpha.0] - 2021-01-27
 
 - move adding and enabling of service descriptions to ServiceController
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -2,7 +2,7 @@
 
 **Technical Specification**
 
-Version: 1.1.8
+Version: 1.1.9
 Doc. ID: XRDSST-CONF
 
 | Date       | Version     | Description                                                                  | Author             |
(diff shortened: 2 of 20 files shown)


bertvi (changes requested):
1. please also update user guide version and version history
2. test_xrdsst.py: please also add integration tests for auto-configuration
3. end_to_end/tests.py: please add tests for query_status and auto-configuration
4. fix linter issues in util.py, api_util.py, test_status.py, test_base_controller.py
5. When running "status" command, everything is filled successfully except the "CSRS" column, that one remains empty
6. test_auto.py unit test seems to test only until cert registration, please also add tests for client and service operations
7. When running "xrdsst apply" command, it runs until it gets an error "Exception calling TokenCertificatesApi->import_certificate: (400)"

unserializable:
>     5. When running "status" command, everything is filled successfully except the "CSRS" column, that one remains empty
Does not reproduce.

>     7. When running "xrdsst apply" command, it runs until it gets an error "Exception calling TokenCertificatesApi->import_certificate: (400)"

Does not reproduce. need more information.

PR #27, opened 2021-02-22
SSTOOLKIT-41

bertvi:


tests/integration/test_xrdsst.py
```diff
@@ -123,6 +123,13 @@ def step_enable_service_description(self, client_id):
             service_description = get_service_description(self.config, client_id)
             assert service_description["disabled"] is False
 
+    def step_add_service_access_rights(self):
+        with XRDSSTTest() as app:
+            service_controller = ServiceController()
+            service_controller.app = app
+            service_controller.load_config = (lambda: self.config)
+            service_controller.add_rights
```
unserializable: Declaration without actual invocation which is not detected by tests; invoke, add application assertions (probably good as another part of assert_server_statuses_transitioned).

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -245,3 +246,4 @@ Further subsystem registration can proceed with ``xrdsst client register``.
 ### 4.9 Service management
 Services and service descriptions are managed with ``xrdsst service`` subcommands. Adding REST/OPENAPI3/WSDL service descriptions
 is performed with ``xrdsst service add-description``. Enabling of service descriptions is performed  with ``xrdsst service enable-description``.
+Adding access rights to services is performed  with ``xrdsst service add-rights``.
```
unserializable: First concrete mention of _services_, earlier has been about _service descriptions_. What differentiates this switch to services here? Also unclear from the paragraph who gets these access rights. Perhaps clarification about services in client context is in order for the section. 

Adding access for 'DEV:GOV:9876:REVIEW' subsystem to use service 'DEV:GOV:9876:REVIEW:Petstore' is trivial, however how to get it across the subsystem, so that  'DEV:GOV:9876:FUNKY' can use 'DEV:GOV:9876:REVIEW:Petstore' ?

Add configuration if needed to cover cross-subsystem usecase (perhaps check out the YAML aliases to understand if they can be fit for purpose).

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.2.4-alpha.0] - 2021-02-22
+- add ``xrdsst service`` subcommand ``add-rights``
```
unserializable: Ambiguity and plural in command name. Remove ambiguity. Some possible variants that come to mind ``add-access``, ``allow-access``, ``enable-access`` (last one would seem to fit together with ``enable-description``) that provides the shared configuration data.

Also remove ``@ex`` label usage, currently ``xrdsst service`` help invocation mentions ('access-rights') instead.

```
xrdsst service
usage: xrdsst service [-h] [-v] {add-description,access-rights,enable-description} ...

Commands for performing service operations.

optional arguments:
  -h, --help            show this help message and exit
  -v, --version         show program's version number and exit

sub-commands:
  {add-description,access-rights,enable-description}
    add-description     Add service description
    access-rights       Add access rights for service
    enable-description  Enable service description
```

requirements.txt
```diff
@@ -20,3 +20,4 @@ jq~=1.1.1
 
 requests~=2.22.0
 python-dateutil~=2.8.1
+pytz~=2019.3
```
unserializable: This dependency seems unnecessary for us, see the comments at ``remote_add_access_rights``, so better remove this, if its only usage would be at tests.

xrdsst/controllers/service.py
```diff
@@ -117,6 +135,38 @@ def remote_enable_service_description(self, ss_configuration, security_server_co
...
+                try:
+                    service_description = self.get_client_service_description(clients_api, client, service_description_conf)
+                    if service_description:
+                        for service in service_description.services:
+                            try:
+                                services_api = ServicesApi(ApiClient(ss_configuration))
+                                service_client = ServiceClient(id=client.id,
+                                                               name=client.member_name,
+                                                               service_client_type=ServiceClientType.SUBSYSTEM,
+                                                               rights_given_at=datetime.now(pytz.utc).isoformat())
```
unserializable: This ServiceClient is data transfer object and it has the ``rights_given_at`` field, however it is not client's responsibility to fill this -- security server can do it just fine. Keep it empty and see about removing the added ``pytz``dependency altogether (@ ``requirements.txt``)

tests/end_to_end/tests.py
```diff
@@ -252,6 +260,7 @@ def test_run_configuration(self):
 
         self.step_add_service_description(client_id)
         self.step_enable_service_description(client_id)
+        self.step_add_service_access_rights()
```
unserializable: Can silently fail, verify transitions afterwards to remedy.

unserializable (changes requested):
Main things:
  * access definitions need clarifications and additions for cross-subsystem, and external subsystem usage and documentation (see notes)
  * state transitions after operations to be verified in tests (integration tests silently pass without performing any access operations, see notes).
  * this seems like it should also be part of autoconfig. If yes, add this operation to dependency operation graph.

tests/integration/test_xrdsst.py
bertvi: Added a check whether service_client has been created, but on my computer this fails because it seems register_client gets HTTP500. Please try if you get the same result. In E2E tests it works at least. It seems that for some reason client registration does not succeed in integration tests and that is the reason why service_client is not added, because it is the same actual client and it need to be registered first.

unserializable:
```python
    def step_add_service_access(self, client_id):
        with XRDSSTTest() as app:
            service_controller = ServiceController()
            service_controller.app = app
            service_controller.load_config = (lambda: self.config)
            service_controller.add_access
```
Integration test fails, the ^^ access invocation is still not done here!

unserializable: So after access invocation was added, the integration test passes on my local dev machine.

xrdsst/core/validator.py
```diff
@@ -188,7 +189,7 @@ def validate_config_client_add_or_register(ss_config, operation, errors):
     return len(errors) <= err_cnt
 
 
-def validate_config_service_desc_add_or_enable(ss_config, operation, errors):
+def validate_config_service_desc(ss_config, operation, errors):
```
unserializable, line 192: Operations are such that they are able to stand on their own and need to validate the configuration for their own only. This validation here has been widened past operation boundaries. Adding / enabling service descriptions is independent of any access definitions and their validation should not be put together with service access validation. The deeper level ``services`` configuration level is also not really required, but striclty optional, right? Lack of it should not fail validation for access definitions.

Also ``service/access`` level currently is not reflected in the error message:

```
AUTO ['service add-access']->'ss5' (redo) 
SKIPPED 'ss5':
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'access' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'services' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'access' missing required value.
```

Where the last level (if applying the required conventions in current code) should be:

```diff
-       'ss5.clients[4].service_descriptions[1]' [service add-access]: 'access' missing required value.
+      'ss5.clients[4].service_descriptions[1].services[1]' [service add-access]: 'access' missing required value.
```

The validation responsibility mixture here means that simple syntactically correct config with no access rights at all will result in having no services services added or enabled, effectively disabling these operations. This is easiest to see during autoconfiguration"

```
AUTO ['service add-description']->'ss5' (redo) 
SKIPPED 'ss5':
        'ss5.clients[4].service_descriptions[1]' [service add-description]: 'access' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-description]: 'services' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-description]: 'access' missing required value.
AUTO ['service add-access']->'ss5' (redo) 
SKIPPED 'ss5':
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'access' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'services' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service add-access]: 'access' missing required value.
AUTO ['service enable-description']->'ss5' (redo) 
SKIPPED 'ss5':
        'ss5.clients[4].service_descriptions[1]' [service enable-description]: 'access' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service enable-description]: 'services' missing required value.
        'ss5.clients[4].service_descriptions[1]' [service enable-description]: 'access' missing required value.
```

Split the validation responsibilities for access grants so that single operations stand on their own.

unserializable (changes requested):

PR #29, opened 2021-02-23
SSTOOLKIT-45

bertvi:


unserializable:
With separate ``update-parameters`` handling, the service description section has a mixture of configuration that is handled in a way that is neither simple to describe nor intuitive to use. If ``ssl_auth=True`` and ``ssl_auth_all=False``, who, if anyone will using SSL? What about ``timeout`` & ``timeout_all``? And having ``url`` and ``url_all`` in a single section does not offer clarity either.

Having functioning setup and then adding a full section:

```
      service_descriptions:
        ...
        - url: <SERVICE_DESCRIPTION_URL>
          rest_service_code: <REST_SERVICE_CODE>
          type: <SERVICE_TYPE>
          timeout: <SERVICE_TIMEOUT>
          ssl_auth: <SERVICE_USE_SSL_AUTH>
          url_all: <SERVICE_URL_FOR_ALL>
          timeout_all: <SERVICE_TIMEOUT_FOR_ALL>
          ssl_auth_all: <SERVICE_USE_SSL_AUTH_FOR_ALL>
          ignore_warnings: <SERVICE_IGNORE_WARNINGS>
        ...

```
followed by ``service add-description`` then does not at all affect parameters that are 'update only'. This diverges from 1->2->3, (change configuration file -> run toolkit -> be done).

The ``ignore_warnings`` is in fact applicable both for service description (schema validation), and single services / service bundles.

These issues seem possible to address only after definining logical separation of service descriptions and services.

PR #35, opened 2021-03-12
Extended proxy error handling and feedback, ...

unserializable:
* Extend the API clients rudimentary API exception at call site
* 'Excplanation' class with mechanisms for interpreting extended API exceptions, for Proxies
* Text maps for proxy errors with error descriptions and some hints for applicable cases mainly at Client Proxy
* Message flow ASCII art to go together with console error messages.
* SOAP and REST/XML failures from producer side that are not converted by server adapters are left unprocessed.

### DIFF
--- a/xrdsst/core/excplanation.py
+++ b/xrdsst/core/excplanation.py
@@ -1,0 +1,125 @@
+# The attempt to provide helpful interpretation for the security server error messages
+# originating from proxies. Main inspiration for this is "How to Interpret Security Server
+# Proxy Error Messages?" page at https://confluence.niis.org/pages/viewpage.action?pageId=53445058
+
+import functools
+import json
+import re
+from json.decoder import JSONDecodeError
+
+from requests.status_codes import codes
+
+from xrdsst.resources.texts import server_error_map, ascii_art, server_hint_map
+from xrdsst.rest.rest import ApiException
+
+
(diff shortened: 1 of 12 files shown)


xrdsst/core/excplanation.py
```diff
@@ -0,0 +1,125 @@
...
+from json.decoder import JSONDecodeError
+
+from requests.status_codes import codes
+
+from xrdsst.resources.texts import server_error_map, ascii_art, server_hint_map
+from xrdsst.rest.rest import ApiException
+
+
+# /Explained Exception/, with current implementation revolving around ApiException interpretations.
+class Excplanatory:
```
bertvi, line 17: I assume the class name "Excplanatory" should be "Explanatory" and also the file name "excplanation" as "explanation"

bertvi (changes requested):

xrdsst/core/excplanation.py
unserializable: Name is as intended, as should be very clear from comments. Other variants considered were ``Excplation`` and ``ExplainingExceptionWrapper``.

bertvi: OK, I did not know that you have come up with a totally new word here. If you think this will not cause any misunderstanding or errors due to writing it differently, then it is not an issue.

bertvi (approved):

PR #42, opened 2021-03-29
SSTOOLKIT-56

bertvi:
Make SSH user for api key creation configurable from configuration file and move api_key section into security_server section in the configuration file.

### DIFF
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -1,5 +1,5 @@
+admin_credentials: <SECURITY_SERVER_CREDENTIALS>
 ssh_access:
-  admin_credentials: <SECURITY_SERVER_CREDENTIALS>
   user: <SSH_USER>
   private_key: /path/to/ssh_private_key
 security_server:
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -1,14 +1,14 @@
 admin_credentials: <SECURITY_SERVER_CREDENTIALS>
+api_key_roles:
+- XROAD_SYSTEM_ADMINISTRATOR
+- XROAD_SERVICE_ADMINISTRATOR
+- XROAD_SECURITY_OFFICER
+- XROAD_REGISTRATION_OFFICER
 ssh_access:
   user: <SSH_USER>
   private_key: /path/to/ssh_private_key
 security_server:
 - api_key: X-Road-apikey token=<API_KEY>
-  api_key_roles:
-  - XROAD_SYSTEM_ADMINISTRATOR
-  - XROAD_SERVICE_ADMINISTRATOR
-  - XROAD_SECURITY_OFFICER
-  - XROAD_REGISTRATION_OFFICER
   api_key_url: https://localhost:4000/api/v1/api-keys
   admin_credentials:
   configuration_anchor: /path/to/configuration-anchor.xml
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -5,7 +5,10 @@ ssh_access:
 security_server:
 - api_key: X-Road-apikey token=<API_KEY>
   api_key_roles:
-    - <SECURITY_SERVER_ROLE_NAME>
+  - XROAD_SYSTEM_ADMINISTRATOR
+  - XROAD_SERVICE_ADMINISTRATOR
+  - XROAD_SECURITY_OFFICER
+  - XROAD_REGISTRATION_OFFICER
   api_key_url: https://localhost:4000/api/v1/api-keys
   admin_credentials:
   configuration_anchor: /path/to/configuration-anchor.xml
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -1,5 +1,5 @@
 ssh_access:
-  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
+  admin_credentials: <SECURITY_SERVER_CREDENTIALS>
   user: <SSH_USER>
   private_key: /path/to/ssh_private_key
 security_server:
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -1,12 +1,13 @@
-security_server:
-- api_key:
-  - key: X-Road-apikey token=<API_KEY>
-    credentials: <SECURITTY_SERVER_CREDENTIALS>
-    ssh_user: <SSH_USER>
-    ssh_key: /path/to/ssh_private_key
-    roles:
+ssh_access:
+  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
+  user: <SSH_USER>
+  private_key: /path/to/ssh_private_key
+security-server:
+- api_key: X-Road-apikey token=<API_KEY>
+  api_key_roles:
     - <SECURITY_SERVER_ROLE_NAME>
-    url: https://localhost:4000/api/v1/api-keys
+  api_key_url: https://localhost:4000/api/v1/api-keys
+  admin_credentials:
   configuration_anchor: /path/to/configuration-anchor.xml
   certificates:
     - /path/to/signcert
@@ -21,6 +22,8 @@ security_server:
   software_token_pin: <SOFT_TOKEN_PIN>
   fqdn: <SECURITY_SERVER_EXTERNAL_FQDN>
   url: https://<SECURITY_SERVER_INTERNAL_FQDN_OR_IP>:4000/api/v1
+  ssh_user: <SSH_USER>
+  ssh_private_key: /path/to/ssh_private_key
   clients:
     - member_class: <MEMBER_CLASS>
       member_code: <MEMBER_CODE>
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -2,7 +2,7 @@ ssh_access:
   admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
   user: <SSH_USER>
   private_key: /path/to/ssh_private_key
-security-server:
+security_server:
 - api_key: X-Road-apikey token=<API_KEY>
   api_key_roles:
     - <SECURITY_SERVER_ROLE_NAME>
--- a/config/xrdsst.yml
+++ b/config/xrdsst.yml
@@ -1,15 +1,16 @@
-api_key:
-- credentials: xrd:secret
-  key: /path/to/ssh_private_key
-  roles:
-  - <SECURITY_SERVER_ROLE_NAME>
-  url: https://localhost:4000/api/v1/api-keys
 security_server:
-- api_key: X-Road-apikey token=<API_KEY>
+- api_key:
+  - key: X-Road-apikey token=<API_KEY>
+    credentials: <SECURITTY_SERVER_CREDENTIALS>
+    ssh_user: <SSH_USER>
+    ssh_key: /path/to/ssh_private_key
+    roles:
+    - <SECURITY_SERVER_ROLE_NAME>
+    url: https://localhost:4000/api/v1/api-keys
+  configuration_anchor: /path/to/configuration-anchor.xml
   certificates:
     - /path/to/signcert
     - /path/to/authcert
-  configuration_anchor: /path/to/configuration-anchor.xml
   name: <SECURITY_SERVER_NAME>
   owner_dn_country: <OWNER_DISTINGUISHED_NAME_COUNTRY>
   owner_dn_org: <OWNER_DISTINGUISHED_NAME_ORGANIZATION>
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -110,14 +110,15 @@ $ pip install setup.py
   
 ### 3.2 Format of configuration file
 ```
-api-key:
-- credentials: <SECURITTY_SERVER_CREDENTIALS>
-  key: /path/to/ssh_private_key
-  roles:
-  - <SECURITY_SERVER_ROLE_NAME>
-  url: https://localhost:4000/api/v1/api-keys
 security-server:
-- api_key: X-Road-apikey token=<API_KEY>
+- api_key:
+  - key: X-Road-apikey token=<API_KEY>
+    credentials: <SECURITTY_SERVER_CREDENTIALS>
+    ssh_user: <SSH_USER>
+    ssh_key: /path/to/ssh_private_key
+    roles:
+    - <SECURITY_SERVER_ROLE_NAME>
+    url: https://localhost:4000/api/v1/api-keys
   configuration_anchor: /path/to/configuration-anchor.xml
   certificates:
     - /path/to/signcert
@@ -160,6 +161,7 @@ The ``logging`` section is for configuring the logging parameters of the X-Road 
 The ``security-server`` section is for configuring security server parameters
 
 * <SECURITY_SERVER_CREDENTIALS> X-Road Security Server credentials, e.g. xrd:secret
+* <SSH_USER> SSH username
 * ``/path/to/ssh_private_key`` should be substituted with the correct path to the ssh private key file, e.g. home/user/id_rsa
 * <SECURITY_SERVER_ROLE_NAME> parameter required for security server api key, should be substituted with a security server role name, e.g. XROAD_SYSTEM_ADMINISTRATOR    
 * ``/path/to/xrdsst.log`` should be substituted with the correct path to the log file, e.g. "/var/log/xroad/xrdsst.log"
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -111,9 +111,10 @@ $ pip install setup.py
 ### 3.2 Format of configuration file
 ```
 ssh_access:
+  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
   user: <SSH_USER>
   private_key: /path/to/ssh_private_key
-security-server:
+security_server:
 - api_key: X-Road-apikey token=<API_KEY>
   api_key_roles:
     - <SECURITY_SERVER_ROLE_NAME>
@@ -158,9 +159,8 @@ security-server:
               url: <SERVICE_URL>
 ```
 
-The ``api-key`` section is for configuring the automatic api key generation parameters for security server
-The ``logging`` section is for configuring the logging parameters of the X-Road Security Server Toolkit
-The ``security-server`` section is for configuring security server parameters
+The ``ssh_access`` section is for configuring the SSH access parameters of the X-Road Security Server Toolkit
+The ``security_server`` section is for configuring security server parameters
 
 * <SECURITY_SERVER_CREDENTIALS> X-Road Security Server admin credentials, e.g. xrd:secret (if specified in ``ssh_access`` section, one value will be 
   used for all configurable security servers, but if specified in the ``security_server`` section, the value will be overridden for specific 
--- a/docs/xroad_security_server_toolkit_user_guide.md
+++ b/docs/xroad_security_server_toolkit_user_guide.md
@@ -2,7 +2,7 @@
 
 **Technical Specification**
 
-Version: 1.2.5
+Version: 1.2.6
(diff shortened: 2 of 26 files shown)


unserializable:
Concept & goal review based on configuration sample and documentation below, no code review at this point.

  * Multiple ssh access keys and API keys per security server can still be specified, which is unlikely to be desirable and also it appears they are not used,  e.g. ``run_end_to_end_tests.sh``  uses only the first definition.
  * The configuration element has not been disambiguated from the ``api_key``. SSH access is multi-purpose and should be separated clearly. The head-aches arising from this are multiple:
    * ``api_key/credentials`` actually defines administrative security server logon
    * ``api_key/key`` is artificial and unintuitive construction.
      * As SSH and API credentials are mixed, differentiation of configurations where the key is created via SSH or preconfigured is unclear.
    * Arising from the previous, the configuration elements that do not have effect at all still linger in the configuration, even if they have no effect, ``roles`` will never be applied when SS API key is preconfigured.
  * Newly added ``ssh_key`` seems better off disambiguated to also emphasize the private parts, e.g.  ``ssh_private_key``
  * There does not seem to be anything wrong with allowing to define SSH access parameters both on root level and security server to accommodate these use-cases with much shorter configuration, so I suggest to take that route to avoid unduly burden and around dozen lines of configuration for each security server that uses automatically created API keys. Moving the SSH access elements solely to the security server means that for scenarios where the SSH access are bound to be identical, like upcoming HA / LB setups, this configuration needs to be repeated for all of the security servers, which is lengthy and error prone, especially the c/p way.

1. See criterias at the task, which are the main focus of first comments above.
2. Design for minimalism and clarity here, go forward to split the configuration elements so that logically unrelated concepts are **not being mixed**. This in turn should also naturally help to get rid of the cases where configuration has lengthy elements that will have no effect at all.
3. Keep the forward-thinking UX in mind and evolve the design so that duplicated configurations are not needed for HA / LB scenarios.

config/xrdsst.yml
```diff
@@ -1,15 +1,17 @@
-api_key:
-- credentials: xrd:secret
-  key: /path/to/ssh_private_key
-  roles:
-  - <SECURITY_SERVER_ROLE_NAME>
-  url: https://localhost:4000/api/v1/api-keys
+ssh_access:
+  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
```
unserializable: This was among those suggested to decouple first from ``api_key`` and then from ``ssh`` too. And more generally, in the high-level review, about "design for minimalism and clarity here, go forward to split the configuration elements so that logically unrelated concepts are not being mixed."

We did not get into benefits though, so lets' consider them more explicitly, decoupling has tangible benefits for not getting caught in entangled configuration to support simple scenarios like:
  * running toolkit SSH-less on the configured security server itself
  * SSH-less operation when remote API calls are enabled

So let's get all the way here in and separate these concerns in the configuration.

Also, ``SECURITTY`` typo can be fixed.

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -27,7 +27,7 @@ Doc. ID: XRDSST-CONF
 | 17.03.2021 | 1.2.3       | Describe failure interpretation and recovery                                 | Taimo Peelo        |
 | 22.03.2021 | 1.2.4       | Default configuration from config/base.yaml -> config/xrdsst.yml             | Taimo Peelo        |
 | 26.03.2021 | 1.2.5       | Add 'fqdn' key for security server, fix service field descriptions           | Taimo Peelo        |
-
+| 29.03.2021 | 1.2.6       | Move api_key section into security server section                            | Bert Viikmäe       |
```
unserializable: Document change note is now outdated and does not describe changes well.

tests/resources/test-config-template.yaml
```diff
@@ -1,20 +1,19 @@
...
-  level: INFO
-security_server:
-- api_key: X-Road-apikey token=<API_KEY>
+  api_key_url: https://localhost:4000/api/v1/api-keys
+  admin_credentials:
   certificates:
-    - /path/to/signcert
-    - /path/to/authcert
+    - /home/bertvi/x/ss3-sign.pem
+    - /home/bertvi/x/ss3-auth.pem
```
unserializable: I do kind of like the added extension here, but unfortunately there are several certificate formats supported, so I guess this template certificate path change is better off reverted,  since ``authcert`` reflects purpose and essence bit better than ``ss3-auth.pem``

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -110,14 +110,16 @@ $ pip install setup.py
...
+ssh_access:
+  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
+  user: <SSH_USER>
+  private_key: /path/to/ssh_private_key
+security_server:
 - api_key: X-Road-apikey token=<API_KEY>
+  api_key_roles:
+    - <SECURITY_SERVER_ROLE_NAME>
+  api_key_url: https://localhost:4000/api/v1/api-keys
+  admin_credentials: <SECURITTY_SERVER_CREDENTIALS>
```
unserializable: Second ``SECURITTY`` typo.

xrdsst/controllers/base.py
```diff
@@ -59,16 +59,19 @@ def _pre_argument_parsing(self):
                                 metavar='file',
                                 default=os.path.join(ROOT_DIR, BaseController._DEFAULT_CONFIG_FILE))
 
-    def create_api_key(self, roles_list, config, security_server):
+    def create_api_key(self, roles_list, security_server):
```
unserializable: The failures should be more expressive here, e.g. following  simple configuration snippet from inited server (works nicely when pre-created API key is filled in):

```yaml
ssh_access:
  admin_credentials: xrd:secret
  user: root
  private_key: /home/user/.ssh/private
security_server:
  api_key_roles:
    - XROAD_SYSTEM_ADMINISTRATOR
    - XROAD_SECURITY_OFFICER
    - XROAD_SERVICE_ADMINISTRATOR
    - XROAD_REGISTRATION_OFFICER
  configuration_anchor: /home/user/path/cs-anchor-internal/anchor.xml
  name: ss5
  .. snip ..
  software_token_id: 0
  software_token_pin: 1122
```

fails without containing any references to user name or other possible problems, from the descriptions I cannot see what it would be missing, this is on a server where "root" has been perfectly valid for login for long time:

```sh
$ xrdsst -c config/newss5.yaml  token list 
Exception calling BaseController->get_api_key:: 'admin_credentials'
Exception when calling TokensApi->get_tokens: (401)
Reason: 
HTTP response headers: HTTPHeaderDict({'x-road-ui-correlation-id': '1242d29ad9f54261', 'X-Content-Type-Options': 'nosniff', 'X-XSS-Protection': '1; mode=block', 'Cache-Control': 'no-cache, no-store, max-age=0, must-revalidate', 'Pragma': 'no-cache', 'Expires': '0', 'Strict-Transport-Security': 'max-age=31536000 ; includeSubDomains', 'X-Frame-Options': 'DENY', 'Content-Security-Policy': "default-src 'none'", 'Content-Type': 'application/json', 'Transfer-Encoding': 'chunked', 'Date': 'Tue, 30 Mar 2021 19:36:16 GMT'})
HTTP response body: {"status":401}
```
The first message comes from the ``create_api_key`` block that catches general Exception, and was marked for refactor in the earlier review. Immediately preceding it is block:

```python
else:
    self.log_api_error('API key creation for security server ' + security_server['name'] + ' failed.')
```
That is similarly quiet about problem source which it supposedly does capture earlier on the connection attempt. Indeed this will even fail with another exception instead of logging, since it calls 2-argument ``self.log_api_error`` with single argument.

Using same configuration with ``token login`` fails similarly with mysterious 'admin_credentials', but since it cannot retrieve status without API KEY, the impression given is that server has not been initialized at all:

```sh
$ xrdsst -c config/newss5.yaml  token login
ERROR:root:Exception calling BaseController->get_api_key:: 'admin_credentials'
Exception calling BaseController->get_api_key:: 'admin_credentials'
SKIPPED 'ss5': needs ['init'] completion before continuing with requested ['token login']
```

In addition to making SSH parameters more explicit to pinpoint unexpected failure, this can beforehand also validate that configuration keys combination needed for SSH access to proceed are actually present and fail early to provide better recovery.

config/xrdsst.yml
```diff
@@ -1,15 +1,17 @@
...
+  user: <SSH_USER>
+  private_key: /path/to/ssh_private_key
 security_server:
 - api_key: X-Road-apikey token=<API_KEY>
+  api_key_roles:
```
unserializable: These ``api_key_roles`` are for /temporary/ keys only, which are anyway created with administrative credentials, so can have any roles possible and indeed should have them pretty much constant unless some new roles appear. How about just dropping it completely and filling in at code? Or at the very least, not requiring their definition at server level, but only at root level (which is still kind of annoying, but at least could be feasibly argued to provide forward-compatibility)?

unserializable (changes requested):

config/xrdsst.yml
bertvi: Fixed typo

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Updated

tests/resources/test-config-template.yaml
bertvi: Fixed

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Fixed

config/xrdsst.yml
bertvi: The roles are now listed and placeholders removed, the user does not need to fill the roles any more

config/xrdsst.yml
bertvi: Moved admin_credentials out from ssh_access section to have it completely separate from SSH.

xrdsst/controllers/base.py
bertvi: Fixed general exception in create_api_key method to have more information related to exit_code and output data.

docs/xroad_security_server_toolkit_user_guide.md
unserializable: Hey, the ``outdated`` was not to be taken literally, though date change is of course fine -- the gist was that the *change description* itself is outdated, the 'moving of the api_key' does reflect on added configuration elements and actual changes.

config/xrdsst.yml
unserializable: Great, listing them explicitly seems very reasonable indeed. However, the following consideration was not clarified:

> How about just dropping it completely and filling in at code? Or at the very least, not requiring their definition at server level, but only at root level (which is still kind of annoying, but at least could be feasibly argued to provide forward-compatibility)?

So what were your thoughts about these two points, were there any particular reasons you preferred to keep these at server level instead of root level? Did you see any downsides to removing the element completely (having slept on it, it seems actually better than in the evening)?

bertvi: It would be exceptionally a bad idea to hard-code them in the source code. I agree with the comment that it is unreasonable for the user to type in these roles, but the roles have to remain in the configuration to be changed if required and also to let the user know what kind of roles the api-key has or must have in order to operate the toolkit.

First you commented that the api-key should be per security server, moved it into the security server, now you are recommending to move the roles out to a separate section or lose completely. Yes, I have given it some thought and I disagree with those comments and remain in the opinion that the roles should be kept where the api-key is and list them there.

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Updated the description of the changes

config/xrdsst.yml
unserializable: > First you commented that the api-key should be per security server, moved it into the security server, now you are recommending to move the roles out to a separate section or lose completely. Yes, I have given it some thought and I disagree with those comments and remain in the opinion that the roles should be kept where the api-key is and list them there.

The ``api_key`` indeed only can be per security server. This is about  ``api_key_roles`` which are for temporary api_keys only and have remained constant. So for persistent API keys used this configuraton does not give any information about API key roles. Both persistent API key and transient API key roles are listed in server status query report.

> It would be exceptionally a bad idea to hard-code them in the source code. I agree with the comment that it is unreasonable for the user to type in these roles, but the roles have to remain in the configuration to be changed if required and also to let the user know what kind of roles the api-key has or must have in order to operate the toolkit.

Could you explain what makes this idea exceptionally bad? I see only forward-compatibility concerns I listed in first comment (which would apply when toolkit becomes unmaintained at one point and X-Road breaks backwards compatibility), but that does not seem exceptionally bad, as the role changes are really slow in X-Road which itself is designed to be backwards-compatible.

xrdsst/controllers/base.py
unserializable: This now indeed passes the Exception message forward, but with the previously cited configuration, where just ``admin_credentials`` is moved to root leve, it now fails as follows:

```sh
$ xrdsst -c config/newss5.yaml token list
Exception calling BaseController->get_api_key:: 'NoneType' object is not iterable
Exception when calling TokensApi->get_tokens: (401)
Reason: 
HTTP response headers: HTTPHeaderDict({'x-road-ui-correlation-id': 'c89999342193ee17', 'X-Content-Type-Options': 'nosniff', 'X-XSS-Protection': '1; mode=block', 'Cache-Control': 'no-cache, no-store, max-age=0, must-revalidate', 'Pragma': 'no-cache', 'Expires': '0', 'Strict-Transport-Security': 'max-age=31536000 ; includeSubDomains', 'X-Frame-Options': 'DENY', 'Content-Security-Policy': "default-src 'none'", 'Content-Type': 'application/json', 'Transfer-Encoding': 'chunked', 'Date': 'Wed, 31 Mar 2021 08:10:02 GMT'})
HTTP response body: {"status":401}

```

config/xrdsst.yml
bertvi: Hard-coded values in the code is generally a bad idea, there is not much to discuss/explain about it. As the roles are related to api keys and only to that in the sense of configuration, these should be kept close to api key parameters, not in some separate section, which would make it confusing

xrdsst/controllers/base.py
bertvi: The current implementation required the admin_credentials, ssh_private_key and ssh_user paramters to still be present in the configuration, but could be left empty, the exception resulted when these were totally missing, now amended the source code to check whether these are totally missing

unserializable: Yup, previously presented configuration snippet now works at console as expected, to have this staked down, add good testcase that tests configuration propagation from configuration root to security servers and their priority, asserting that values that end up being used for ``ssh`` are the same as expected/documented.

bertvi: Added 6 unit tests for testing retrieval of admin_credentials and SSH parameters from different configuration file sections

unserializable: Tests seem okay at first glance.

Further manual testing shows that configured key is not the one solely used, e.g. creating brand new keypair ``justnewkey`` and robustly substituting it into configuration file:

```diff
-private_key: /home/user/.ssh/real
+private_key: /home/user/.ssh/justnewkey
```

still allows API key creation, if only few keys are present under ``~/.ssh``, since SSH makes attempts to use other keys under  there, which can either succeed when the amount of keys are small or fails with ``Too many authentication failures`` when there are more keys and the one that allows access to server is not reached 'fast enough'.

It seems it is better to fail at once there, instead of attempting fallbacks, so that misconfigurations are revealed earlier.

config/xrdsst.yml
unserializable: There is much do discuss here, if not blankets, but the case.

So the current standing for specifying transient key roles for 3 security servers, ``as-is-keyroles.yml``

```yaml
# Each temporary SSH key has explicit roles listed at each security server,
# these are available for temporary keys via the administrative credentials.
# If any of these are missing, not all toolkit operations are invokable.
admin_credentials: admuser:admpass
ssh_access:
  user: some_user
  private_key: /path/to/ssh_private_key
security_server:
- name: ss1
  url: https://ss1:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_roles:
  - XROAD_SYSTEM_ADMINISTRATOR
  - XROAD_SERVICE_ADMINISTRATOR
  - XROAD_SECURITY_OFFICER
  - XROAD_REGISTRATION_OFFICER
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss1 configuration
- name: ss2
  url: https://ss2:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_roles:
  - XROAD_SYSTEM_ADMINISTRATOR
  - XROAD_SERVICE_ADMINISTRATOR
  - XROAD_SECURITY_OFFICER
  - XROAD_REGISTRATION_OFFICER
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss2 configuration
- name: ss3
  url: https://ss3:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_roles:
  - XROAD_SYSTEM_ADMINISTRATOR
  - XROAD_SERVICE_ADMINISTRATOR
  - XROAD_SECURITY_OFFICER
  - XROAD_REGISTRATION_OFFICER
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss3 configuration
```

When transient key roles were being defined in application and entirely left out of the configuration, ``code-applied-keyroles.yml``

```yaml
# Roles needed by toolkit are added to temporary key from code,
# these are available for temporary keys via the administrative credentials.
# All toolkit operations are invokable.
admin_credentials: admuser:admpass
ssh_access:
  user: some_user
  private_key: /path/to/ssh_private_key
security_server:
- name: ss1
  url: https://ss1:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss1 configuration
- name: ss2
  url: https://ss2:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss2 configuration
- name: ss3
  url: https://ss3:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss3 configuration
```

And finally, transient key roles defined in the configuration, but only at root level, ``root-keyroles.yml``

```yaml
# Temporary SSH keys have explicit roles listed at root of configuration file,
# again available for temporary keys via the administrative credentials.
# If any of these are missing, not all toolkit operations are invokable.
admin_credentials: admuser:admpass
ssh_access:
  user: some_user
  private_key: /path/to/ssh_private_key
api_key_roles:
  - XROAD_SYSTEM_ADMINISTRATOR
  - XROAD_SERVICE_ADMINISTRATOR
  - XROAD_SECURITY_OFFICER
  - XROAD_REGISTRATION_OFFICER
security_server:
- name: ss1
  url: https://ss1:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss1 configuration
- name: ss2
  url: https://ss2:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss2 configuration
- name: ss3
  url: https://ss3:4000/api/v1
  api_key: X-Road-apikey token=<API_KEY>
  api_key_url: https://localhost:4000/api/v1/api-keys
  # OTHER ss3 configuration
```

So as the key is transient and roles are backwards compatible, after considering this and taking note that ``api_key_roles`` element would most natural next to API keys, I now suggest we go with the code applied keyroles entirely (before I was leaning towards root definition for key roles), as the side effects seem to be strictly about forward-compatibilty, which we can safely ignore, as toolkit code can always take care of them, if these occur in the future.

bertvi: I moved the api_key_roles to the root level

PR #43, opened 2021-03-30
Insensitive logs sstoolkit 66

unserializable:
Masks the API key representation in log files, revealing ~12 bits of entropy from original ~120 bits of entropy.

```
7b6b39d5-64a4-428b-a0b3-ed950acab9bf  # original
********-****-428b-****-************  # masked
```

Side-quests:

Restore the no-log-configured temporary log file creation behaviour for single operations, bisected to f28fd3bd9574c3acafc451aadd8012c7ea51bada borkage where ``init_logging(...)`` gets applied **after** first status queries, for which unconfigured API_KEY scenario invokes ``get_api_key(..) -> create_api_key(...)`` which logging attempts create StreamHandler with unset log level and then kept once-only application of init_logging() from occuring, since this activated only when /no/ log handlers existed yet.

### DIFF
--- a/xrdsst/controllers/base.py
+++ b/xrdsst/controllers/base.py
@@ -3,6 +3,7 @@ import copy
 import functools
 import inspect
 import json
+import re
 import sys
 
 import networkx
@@ -35,6 +36,14 @@ BANNER = texts['app.description'] + ' ' + get_version() + '\n' + get_version_ban
 class BaseController(Controller):
     _TRANSIENT_API_KEY_ROLES = ['XROAD_SYSTEM_ADMINISTRATOR', 'XROAD_SERVICE_ADMINISTRATOR', 'XROAD_SECURITY_OFFICER', 'XROAD_REGISTRATION_OFFICER']
     _DEFAULT_CONFIG_FILE = "config/xrdsst.yml"
+    _RE_API_KEY = re.compile(r"""
+        ([a-f0-9]{8}-
+        [a-f0-9]{4}-)
+        ([a-f0-9]{4})-  # UUID version + 3 hexdecimal digits, only part retained
+        ([a-f0-9]{4}-  # Do no validate first character separately
+        [a-f0-9]{12})
+    """, re.VERBOSE | re.IGNORECASE)
+
     class Meta:
         label = 'base'
         stacked_on = 'base'
@@ -42,6 +51,11 @@ class BaseController(Controller):
         arguments = [
             (['-v', '--version'], {'action': 'version', 'version': BANNER})
         ]
+
+    @staticmethod
+    def api_key_scrambler(log_record):
+        log_record.msg = re.sub(BaseController._RE_API_KEY, '********-****-\\2-****-************', str(log_record.msg))  # clear ~108 bits from ~120
+        return 1
 
     @staticmethod
     def get_server_status(api_config, ss_config):
@@ -239,10 +253,13 @@ class BaseController(Controller):
 
     @staticmethod
     def init_logging(configuration):
-        if logging.getLogger().handlers:
-            return
+        curr_handlers = logging.getLogger().handlers
+        if curr_handlers:
+            if any(map(lambda h: h.level != logging.NOTSET, curr_handlers)):  # Skip init ONLY if ANY handler levels set.
+                return
 
         exit_messages = ['']
+        logging.getLogger().handlers = []
 
         auto_log_file_name = str(Path.home()) + "/" + texts['app.label'] + "-" + \
                              datetime.now().strftime("%Y%m%d-%H%M-%S") + \
@@ -277,6 +294,10 @@ class BaseController(Controller):
                 logging.basicConfig(filename=log_file_name, level=log_level, format=log_format)
             exit_messages.append("Activities logged into '" + log_file_name + "'.")
             atexit.register(lambda: print(*exit_messages, sep='\n'))
+
+        for handler in logging.getLogger().handlers:
+            handler.addFilter(BaseController.api_key_scrambler)
+            handler.setLevel(log_level)
 
     def load_config(self, baseconfig=None):
         # Add errors to /dict_err_lists/ at given key for /sec_server_configs/ that do not have required /key/ defined.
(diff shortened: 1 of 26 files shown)


xrdsst/controllers/base.py
```diff
@@ -34,6 +35,14 @@
 
 class BaseController(Controller):
     _DEFAULT_CONFIG_FILE = "config/xrdsst.yml"
+    _RE_API_KEY = re.compile(r"""
```
bertvi, line 39: This would better fit into the util.py where there is already "RE_API_KEY_HEADER"

bertvi (changes requested):
For some reason I only see "INFO - API key '********-****-4f10-****-************' for security server ss3 revoked" messages in the logs, but not "API key "6e8c4a69-6b4c-4279-b862-176e54b2e19f" for security server ss3 created" messages, although these should also be logged at INFO level. If only revoke is logged, it is confusing and also creation of the api key should be logged.

xrdsst/controllers/base.py
unserializable: Yes, this code location was considered first, however it was put into the BaseController and marked with Python's largely useless 'protected' convention as it is directly used in output sanitation and safer out of ``core/util``, and  other packages where eager development happens more often.

However, there is a simple alternative, getting rid of the ``core/util.py`` ``RE_API_KEY_HEADER`` entirely by requiring configuration API key definition to match the one returned by the security server, solely relying on the 'protected' field of ``BaseController``, as clearly even the presence of so similar regular expressions is ``code smell``.

AFAIU, so far you have been against this simple change, but from the above it is maybe clear that this is best way to proceed, so please confirm your agreement here.

unserializable:
> For some reason I only see "INFO - API key '********-****-4f10-****-************' for security server ss3 revoked" messages in the logs, but not "API key "6e8c4a69-6b4c-4279-b862-176e54b2e19f" for security server ss3 created" messages, although these should also be logged at INFO level. If only revoke is logged, it is confusing and also creation of the api key should be logged.

@bertvi The reason is simply that logging is activated after we reach the operation with potential for server state mutation (operations that have no such potential are not attempted to be logged). At the API creation time it is not yet known which operation it will be used for, so the creation has not been logged.

So we can go either go with designing bit of delayed logging or start logging of non-mutable operations, which of these avenues are to be preferred?

xrdsst/controllers/base.py
bertvi: Can not get rid of core/util.py as there are other functions/methods inside used also in other places in addition to the BaseController. My suggestion is: either move "RE_API_KEY_HEADER" also to BaseController or move "_RE_API_KEY" to core/util.py. If these are in the same place, it is more clear.

bertvi:
> > For some reason I only see "INFO - API key '********-****-4f10-****-************' for security server ss3 revoked" messages in the logs, but not "API key "6e8c4a69-6b4c-4279-b862-176e54b2e19f" for security server ss3 created" messages, although these should also be logged at INFO level. If only revoke is logged, it is confusing and also creation of the api key should be logged.
> 
> @bertvi The reason is simply that logging is activated after we reach the operation with potential for server state mutation (operations that have no such potential are not attempted to be logged). At the API creation time it is not yet known which operation it will be used for, so the creation has not been logged.
> 
> So we can go either go with designing bit of delayed logging or start logging of non-mutable operations, which of these avenues are to be preferred?

If think it would be best to log from the start so also api key creation.

xrdsst/controllers/base.py
unserializable: > Can not get rid of core/util.py as there are other functions/methods inside used also in other places in addition to the BaseController. My suggestion is: either move "RE_API_KEY_HEADER" also to BaseController or move "_RE_API_KEY" to core/util.py. If these are in the same place, it is more clear.

Here you have misread my quote above:

>> However, there is a simple alternative, getting rid of the `core/util.py` `RE_API_KEY_HEADER` entirely by requiring configuration API key definition to match the one returned by the security server, solely relying on the 'protected' field of `BaseController`, as clearly even the presence of so similar regular expressions is `code smell`.

Removal was strictly about field in ``util.py`` and not about ``util.py`` itself). Since you did not answer the question about simply eliminating different forms of API keys above, I will ask once again, with the code sample according to latest suggestion:

Do you indeed still see any reason to have code like:

```python
class BaseController(Controller):
    _RE_API_KEY = re.compile(r"""
        ([a-f0-9]{8}-
        [a-f0-9]{4}-)
        ([a-f0-9]{4})-  # UUID version + 3 hexdecimal digits, only part retained
        ([a-f0-9]{4}-  # Do no validate first character separately
        [a-f0-9]{12})
    """, re.VERBOSE | re.IGNORECASE)

    # Regex of X-Road security server header for API key.
    RE_API_KEY_HEADER = re.compile(r"""
        ^
        X-Road-apikey[ ]token=
        [a-f0-9]{8}-
        [a-f0-9]{4}-
        [a-f0-9]{4}-  # Do not fix UUID version
        [a-f0-9]{4}-  # Do no validate first character separately
        [a-f0-9]{12}
        $
    """, re.VERBOSE | re.IGNORECASE)
```

when it simple and clear to just have  /single/ regular expression by having configuration API key definition match the one returned by the security server?

bertvi: I suppose if your refacto the code so that api-key in the configuration is kept without the "X-Road-apikey token=" you could easily remove the "RE_API_KEY_HEADER".

unserializable: Singe RE retained, https://github.com/nordic-institute/X-Road-Security-Server-toolkit/pull/43/commits/5eb996d0ee34135c33119a724893ff214846fb83

unserializable:
> > So we can go either go with designing bit of delayed logging or start logging of non-mutable operations, which of these avenues are to be preferred?
> 
> If think it would be best to log from the start so also api key creation.

https://github.com/nordic-institute/X-Road-Security-Server-toolkit/pull/43/commits/52578759a70f6fc94ee3d3e0617ec48f39bb097c

bertvi (approved):

PR #45, opened 2021-04-06
SSTOOLKIT-19

bertvi:


create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+  usage
+  exit 1
+}
+
+create_user() {
+    user=$1
+    pass=$2
+    host=$3
+    ssh_key=$4
+    os_name=$(ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "grep '^NAME' /etc/os-release")
```
unserializable: Here, using ``root`` user exclusively does not seem to be best option. Several distros have had or continue to have root login via ``ssh`` disabled by default. IIRC one of the earlier CentOS versions had this disabled by default, though CentOS 8 seems to have it enabled again. Ubuntu 20.04 default installation has ``root`` ssh access disabled as well, see e.g. https://blog.eldernode.com/enable-root-login-via-ssh-in-ubuntu-20-04/

So the approach taken here should probably be flexible and use whatever user account is provided for ssh. Then certain distinctions can be made on how to continue, e.g. ATM I can think of following:
1. if user effective uuid is determined to be 0, then required commands can be executed as if we were root
2. otherwise, if available use the ``sudo`` to execute the required commands. if that fails, go to next step.
3. ... ? something else ? ...
4. ... if nothing of the previous holds or subcommands fail, then report error due to insufficient privileges / command failure.

Good idea there would also be to keep stack of undo commands for successfully executed commands -- in case of process failing mid-way (due to some weird permissions of fine-grained ``sudo`` configuration), the previously executed commands could then be rolled back.

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+
+    # create user and add to sudo/root group
+    if echo "$os_name" | grep 'Ubuntu'; then
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "useradd $user --group sudo -p $pass"
```
unserializable: Why is the ``sudo`` group added here, is it somehow mandatory for the xroad security server administrator user to have ``sudo`` access for some reason on Ubuntu? At the [SS-UG](https://github.com/nordic-institute/X-Road/blob/8d8ac1066a2ee82e0d6cc2a4d9cb54e38226b789/doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md#22-managing-the-users) ``sudo`` access is not mentioned.

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+    fi
+    if echo "$os_name" | grep 'Centos'; then
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "useradd --system --home /var/lib/xroad --no-create-home --shell /bin/bash --user-group --comment \"X-Road system user\" xroad"
```
unserializable: This ``useradd`` seems fragile here, just two failure cases:

1. security server software has ALREADY been installed on Centos, the ``xroad`` user already exists, command will  fail.
2. security server software has NOT been installed on Centos. In such case the user will get **non-existant** home directory.

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "mkdir /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chown xroad:xroad /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chmod 751 /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "touch /etc/xroad/db.properties"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chown xroad:xroad /etc/xroad/db.properties"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chmod 640 /etc/xroad/db.properties"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "useradd $user --group root -p $pass"
```
unserializable: Another user group question arises here -- why is ``$user`` added to the ``root`` group on Centos?

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "mkdir /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chown xroad:xroad /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chmod 751 /etc/xroad"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "touch /etc/xroad/db.properties"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chown xroad:xroad /etc/xroad/db.properties"
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "chmod 640 /etc/xroad/db.properties"
```
unserializable: Lines 35-40: where do all these commands come from, are they from the security server installation script or somewhere else? In any case they seem to attain for the case where no security server package is yet installed at all. But then, the later addition to xroad security server specific groups (line 50) would also be bound to fail, simply because the groups do not exist yet?

This seems to me it is either doing too much with these xroad specific file hierarchies, or too little -- when the security server software indeed has not yet been installed. How would the installer behave when finding these files and folders present, is it able to continue with the correct installation?

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "useradd $user --group root -p $pass"
+    fi
+
+    # add the created user to its own group
+    ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "usermod -a -G $user $user"
+
+    # add the created used to X-Road related groups
+    declare -a StringArray=("xroad-security-officer" "xroad-registration-officer" "xroad-service-administrator" "xroad-system-administrator" "xroad-securityserver-observer")
```
unserializable: This would be better named ``xroad_roles_array``or similar instead of ``StringArray``.

create_admin_user.sh
```diff
@@ -0,0 +1,79 @@
...
+    for group_name in "${StringArray[@]}"; do
+      ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "$ssh_key" root@"$host" "usermod -a -G $group_name $user"
```
unserializable: This assumes existance of user groups listed in ``StringArray`` and created for security server -- which is counterintuitive to parts of earlier script that seem to try to get around even when security server software is not yet installed. If that is indeed **meant** to work in such scenario, the groups would also need to be created, no?

unserializable (changes requested):

create_admin_user.sh
bertvi: "root" user is not used anymore, ssh user is taken from configuration and "sudo" is used within commands. When testing locally it worked for CentOS 7 and Ubuntu18. Could probably be tested for other versions as well.

create_admin_user.sh
bertvi: Removed this part, as it is unneccessary.

create_admin_user.sh
bertvi: It is now checked if these procedures have already been performed or not, so this should be ok

create_admin_user.sh
bertvi: Removed that part

create_admin_user.sh
bertvi: All non-existing groups are now created and this is taken from chapter 2.5.1 in: https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md

create_admin_user.sh
bertvi: This shell script is not used anymore

create_admin_user.sh
bertvi: This script is not used anymore.

create_admin_user.sh
unserializable: wordsfail

^^ 
I still have the same impression -- it is either doing too much or too little. Please explain what it is that is actually being done here, and most importantly, why that is necessary for the user story, what makes _optional_ step 2.5.1 titled _customizing the database properties_ from cited manual even needed here?

setup.cfg
```diff
@@ -18,6 +18,10 @@ ignore-imports = yes
 [DESIGN]
 max-parents = 10
 
+[coverage:run]
+omit =
```
unserializable: Keep in coverage.

xrdsst/controllers/user.py
```diff
@@ -0,0 +1,195 @@
...
+                exitcode, data = subprocess.getstatusoutput(os_name_cmd)
+                if exitcode == 0:
+                    if data.find("Ubuntu") > 0:
+                        return "Ubuntu"
+                    elif data.find("Cent") > 0:
+                        return "Centos"
+                    else:
+                        return None
+                else:
+                    self.log_api_error('UserController->create_user:', 'Determination of operating system for ' \
```
unserializable: Do not use ``log_api_error``, that is really for logging **API** invocation errors, only in very special cases falling through to arguments given.

xrdsst/controllers/user.py
```diff
@@ -0,0 +1,195 @@
...
+            check_xroad_user_cmd = 'ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "{}" {}@{} "{}"'.format(
+                ssh_key, ssh_user, self.security_server_address(security_server), cmd)
+            exitcode, data = subprocess.getstatusoutput(check_xroad_user_cmd)
+            if exitcode == 0:
+                prepare_os = False if len(data) > 0 else True
+            else:
+                self.log_api_error('UserController->create_user:', 'Preparation procedures for adding new user on Centos for ' \
+                                   + security_server['name'] + ' failed (exit_code =' + str(exitcode) + ', data =' + str(data))
+
+            if prepare_os:
```
unserializable: So this is an attempt to execute N commands, with any of them failing and continuing despite failures. What makes this necessary, what creates the confidence that this can achieve known **good** state?

If somebody in Slack can confirm that some or parts of the juggling are necessary then this must be moved into legible script/function to be executed via SSH.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [0.4.1-alpha.0] - 2021-04-06
+- add ``xrdsst user`` and subcommand ``create``
```
unserializable: Perhaps simply``xrdsst create-admin`` if no more user operations are planned to be added or ``xrdsst user create-admin`` if more are planned?

xrdsst/controllers/user.py
```diff
@@ -0,0 +1,195 @@
...
+    @ex(help="Create admin user", arguments=[])
+    def create(self):
+        self.create_user(self.load_config())
+
+    def create_user(self, conf):
+        for security_server in conf["security_server"]:
+            self.log_debug('Creating admin user for security server: ' + security_server['name'])
+            admin_credentials = get_admin_credentials(security_server, conf)
+            user_name = admin_credentials.split(":")[0]
+            pwd = admin_credentials.split(":")[1]
```
unserializable: ```user_name, pwd = admin_credentials.split(":")```  # DRYER

xrdsst/controllers/user.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def determine_os(self, ssh_key, ssh_user, security_server):
+        self.log_debug('Determining operation system for security server: ' + security_server['name'])
+        os_name_cmd = 'ssh -o IdentitiesOnly=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -o LogLevel=ERROR -i "{}" {}@{} "{}"'.format(
+            ssh_key, ssh_user, self.security_server_address(security_server), "grep '^NAME' /etc/os-release")
+        if os.path.isfile(ssh_key):
+            try:
+                exitcode, data = subprocess.getstatusoutput(os_name_cmd)
+                if exitcode == 0:
+                    if data.find("Ubuntu") > 0:
```
unserializable: Arent these comparisons better off case insensitive?

unserializable (changes requested):

xrdsst/controllers/user.py
bertvi: Not using log_api_error() anymore.

CHANGELOG.md
bertvi: Changed the sub-command to create-admin as this will be the only user to be created

xrdsst/controllers/user.py
bertvi: Refactored this sentence

xrdsst/controllers/user.py
bertvi: Made the comparisons case insensitive

create_admin_user.sh
bertvi: For RHEL, the normal way to add the admin user is after installing the packages described in: https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md#28-package-installation

The call described in "2.5.1 Customize the Database Properties" is used if the user needs to be created before the security-server packages have been installed (because the script xroad-add-admin-user comes with the installation) and installation of these packages is out of the scope of this user story

xrdsst/controllers/user.py
bertvi: Amended the code, now in the case of non-zero exit_code, an exception is thrown and the code stops running

create_admin_user.sh
bertvi: This is resolved as "root" login is not performed, the user is taken from "ssh_user" parameter from config file.

CHANGELOG.md
unserializable: This does not seem to match with code description.

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -256,44 +265,63 @@ $ xrdsst status
 ### 4.1 The single command fully automatic configuration of security servers listed in configuration file
 
 The whole security server configuration in a fully automatic mode (all configuration from configuration file) can be run with ``xrdsst apply``
-For performing the configuration step by step instead, please start from [4.2 Initializing the security server](#42-initializing-the-security-server)
+For performing the configuration step by step instead, please start from [4.3 Initializing the security server](#43-initializing-the-security-server)
+
+### 4.2 Creating admin user (optional)
+
+X-Road admin user can be created with ``xrdsst user create``
```
unserializable: Does not match latest code.

unserializable (changes requested):

xrdsst/controllers/user.py
```diff
@@ -0,0 +1,69 @@
...
+        stacked_type = 'nested'
+        description = texts['user.controller.description']
+
+    @ex(help="Create admin user", arguments=[])
+    def create_admin(self):
+        self.create_user(self.load_config())
+
+    def create_user(self, conf):
+        for security_server in conf["security_server"]:
+            self.log_debug('Creating admin user for security server: ' + security_server['name'])
```
unserializable, line 28: Here, only attempt user creation when administrator credentials are present, currently it then fails ungracefully:

```
$ xrdsst -c config/worknewss5.yaml user create-admin 
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/env/bin/xrdsst", line 11, in <module>
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/main.py", line 237, in main
    app.run()
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/core/foundation.py", line 916, in run
    return_val = self.controller._dispatch()
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/env/lib/python3.8/site-packages/cement/ext/ext_argparse.py", line 808, in _dispatch
    return func()
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/controllers/user.py", line 24, in create_admin
    self.create_user(self.load_config())
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/controllers/user.py", line 29, in create_user
    admin_credentials = get_admin_credentials(security_server, conf)
  File "/home/uns/dev/niis/xrdsst-review/X-Road-Security-Server-toolkit/xrdsst/core/util.py", line 8, in get_admin_credentials
    admin_credentials = security_server["admin_credentials"] if security_server.get("admin_credentials", "") else config["admin_credentials"]
KeyError: 'admin_credentials'
```

as it is outside operation graph and has no required/validated inputs defined.

unserializable (changes requested):
Here, do not attempt to create user if no credentials are available.

xrdsst/controllers/user.py
bertvi: Will now fail with exception if admin_credentials, ssh_key or ssh_user is missing as these are required parameters

PR #54, opened 2021-04-26
SSTOOLKIT-69

bertvi:


docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -32,7 +32,7 @@ Doc. ID: XRDSST-CONF
 | 06.04.2021 | 1.2.10      | Notes on user management                                                     | Bert Viikmäe       |
 | 09.04.2021 | 1.2.11      | Added description about signing and verification of packages                 | Bert Viikmäe       |
 | 20.04.2021 | 1.3.0       | Substituting plain text secrets in configuration with environment variables  | Bert Viikmäe       |
-
+| 26.04.2021 | 1.3.1       | Added description about adding endpoints to the REST and OpenAPI services.
```
bertvi: Please also add your name to this table like in the previous rows

tests/end_to_end/tests.py
```diff
@@ -249,6 +250,21 @@ def step_autoconf(self):
...
+        endpoint_controller = EndpointController()
+        for security_server in self.config["security_server"]:
+            configuration = endpoint_controller.create_api_config(security_server, self.config)
+            for client in security_server["clients"]:
+                for service_description in client["service_descriptions"]:
+                    for service in service_description["services"]:
+                        for endpoint in service[0]["endpoints"]:
+                            endpoint_controller.remote_add_service_endpoints(configuration, security_server, client, service_description, endpoint)
+        description = get_service_description(self.config, client_id)
+        assert len(description["services"][0]["endpoints"]) == 1
```
bertvi: This should be "assert len(description["services"][0]["endpoints"]) == 5" as there are 4 generated endpoints in addition to the one you are adding as this description is a real OpenAPI3 description with some endpoints to be generated

tests/end_to_end/tests.py
```diff
@@ -249,6 +250,21 @@ def step_autoconf(self):
...
+
+        assert str(description["services"][0]["endpoints"][0]["path"]) == "/testPath"
```
bertvi: These 2 lines should be: "assert str(description["services"][0]["endpoints"][4]["path"]) == "/testPath" " and
        "assert str(description["services"][0]["endpoints"][4]["method"]) == "POST" " as there are 4 other generated endpoints which are with indexes 0 to 3 in this list. In addition the method to be asserted here should actually be "POST" not "GET", as you have placed "POST" in the test configuration yaml file.

tests/end_to_end/tests.py
```diff
@@ -249,6 +250,21 @@ def step_autoconf(self):
                 auto_controller.app = app
                 auto_controller._default()
 
+    def step_add_service_endpoints(self, client_id):
+        endpoint_controller = EndpointController()
+        for security_server in self.config["security_server"]:
+            configuration = endpoint_controller.create_api_config(security_server, self.config)
+            for client in security_server["clients"]:
+                for service_description in client["service_descriptions"]:
+                    for service in service_description["services"]:
```
bertvi: Instead of this line you should use "for endpoint in service_description["endpoints"]:" as in the configuration file you have placed "endpoints" section inside "service_descriptions" section not in inside "services". The "services" loop you do not need here at all.

tests/integration/test_xrdsst.py
```diff
@@ -146,6 +147,15 @@ def step_update_service_parameters(self, client_id):
...
             assert description["services"][0]["url"] == 'http://petstore.xxx'
 
+    def step_add_service_endpoints(self, client_id):
+        with XRDSSTTest() as app:
+            endpoint_controller = EndpointController()
+            endpoint_controller.app = app
+            endpoint_controller.load_config = (lambda: self.config)
+            endpoint_controller.add_endpoints()
+            description = get_service_description(self.config, client_id)
+            assert len(description["services"][0]["endpoints"]) == 1
```
bertvi: The same as in end to end tests, as this OPENAPI3 description also generated 4 endpoints, you should add the following assertions instead:

            assert len(description["services"][0]["endpoints"]) == 5
            assert str(description["services"][0]["endpoints"][4]["path"]) == "/testPath"
            assert str(description["services"][0]["endpoints"][4]["method"]) == "POST"

tests/resources/test-config-template.yaml
```diff
@@ -11,7 +11,7 @@ security_server:
     - /path/to/authcert
   configuration_anchor: /path/to/configuration-anchor.xml
   name: <SECURITY_SERVER_NAME>
-  owner_dn_country: FI
+  owner_dn_country: EE
```
bertvi: Please revert this change, because in order for these tests to successfully run on Jenkins, this owner_dn_country needs to be "FI".

tests/unit/test_endpoint.py
```diff
@@ -0,0 +1,153 @@
...
+                            sys.stderr.write(err)
+
+    @pytest.fixture(autouse=True)
+    def capsys(self, capsys):
+        self.capsys = capsys
+
+    def test_endpoint_already_added(self):
+        class AlreadyEnabledResponse:
+            status = 409
+            data = '{"status":409,"error":{"code":"service_description_already_enabled"}}'
```
bertvi: substitute with "data = '{"status":409,"error": {"code":"service_endpoint_already_enabled"}}'"

tests/unit/test_service.py
```diff
@@ -298,7 +298,7 @@ def test_service_description_enable(self):
     def test_service_description_enable_already_enabled(self):
         class AlreadyEnabledResponse:
             status = 409
-            data = '{"status":409,"error":{"code":"service_description_already_enabled"}}'
+            data = '{"status":409,"error":{"code":"service_endpoint_already_added"}}'
```
bertvi: Please revert this change as when adding service descriptions, this "service_endpoint_already_added" does not suit here

xrdsst/controllers/endpoint.py
```diff
@@ -0,0 +1,83 @@
...
+                            endpoint = Endpoint(id=None, service_code=service_description_conf["rest_service_code"], method=endpoint_conf["method"], path=endpoint_conf["path"], generated= None)
+
+                            try:
+                                services_api = ServicesApi(ApiClient(ss_api_config))
+                                response = services_api.add_endpoint(id=serviceId, body=endpoint)
+                                if response:
+                                    BaseController.log_info("Added service endpoint '" + endpoint.method + " " + endpoint.path + "' for service '" + serviceId + "'")
+                            except ApiException as err:
+                                if err.status == 409:
+                                    BaseController.log_info("Added service endpoint '" + endpoint.method + " " + endpoint.path  + "' for service '" + serviceId  + "' already added")
```
bertvi: Better use here: "BaseController.log_info("Service endpoint '" + endpoint.method + " " + endpoint.path  + "' for service '" + serviceId  + "' already added")"

xrdsst/controllers/endpoint.py
```diff
@@ -0,0 +1,83 @@
...
+        BaseController.log_keyless_servers(ss_api_conf_tuple)
+
+    def remote_add_service_endpoints(self, ss_api_config, security_server_conf, client_conf, service_description_conf, endpoint_conf):
+        clients_api = ClientsApi(ApiClient(ss_api_config))
+        try:
+            client_controller = ClientController()
+            client = client_controller.find_client(clients_api, security_server_conf, client_conf)
+            if client:
+                try:
+                    service_description = self.get_client_service_description(clients_api, client, service_description_conf)
```
bertvi: Instead of duplicating "get_client_service_description()", remove the method from "endpoint.py" and call "get_client_service_description()" from ServiceController

PR #58, opened 2021-05-04
SSTOOLKIT-70, SSTOOLKIT-82

bertofl90:


config/xrdsst.yml
```diff
@@ -1,48 +1,55 @@
-admin_credentials: <SECURITY_SERVER_CREDENTIALS_OS_ENV_VAR_NAME>
```
bertvi: Please remove your local values from xrdsst.yml and keep the placeholders as they were before and if you add new section, please also add placeholders there instead of real values

run_end_to_end_tests.sh
```diff
@@ -58,6 +58,8 @@ update_config() {
 }
 
 run_tests() {
+  echo "aaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
bertvi: Please remove lines 61 and 62 from run_end_to_end_tests.sh, these seem like some debugging values

tests/end_to_end/tests.py
```diff
@@ -37,7 +37,7 @@ def setUp(self):
                     self.config_file = sys.argv[idx]
             base = BaseController()
             base.app = app
-            self.config = base.load_config(baseconfig=self.config_file)
+            self.config = base.load_config(baseconfig=os.path.join("/home/alberto/Proyects/X-Road-Security-Server-toolkit/", "tests/resources/test-config2.yaml"))
```
bertvi: This change is probably for local cebugging only, so please revert to the previous value.

tests/end_to_end/tests.py
```diff
@@ -141,7 +141,8 @@ def apply_cert_config(self, signed_certs):
         self.config['security_server'][0]['certificates'] = signed_certs
 
     def create_api_key(self, api_key):
-        self.config["security_server"][0]["api_key"] = api_key
+        os.environ["TEST_API_KEY"] = api_key
```
bertvi: Please fetch the changes from master for lines 145-146

tests/integration/test_xrdsst.py
```diff
@@ -158,6 +158,16 @@ def step_add_service_endpoints(self, client_id):
             assert str(description["services"][0]["endpoints"][4]["path"]) == "/testPath"
             assert str(description["services"][0]["endpoints"][4]["method"]) == "POST"
 
+    def step_add_endpoints_access(self, client_id):
```
bertvi, line 161: Please also call your test method "step_add_endpoints_access()" in test_run_configuration in test_xrdsst.py, currently the test step is not run

tests/unit/test_endpoint.py
```diff
@@ -2,12 +2,12 @@
...
-
+import copy
 import pytest
 
 from tests.util.test_util import StatusTestData
 from xrdsst.controllers.endpoint import EndpointController
-from xrdsst.models import Client, ConnectionType, ClientStatus, ServiceDescription, ServiceType, ServiceClient, ServiceClientType, Service
+from xrdsst.models import Client, ConnectionType, ClientStatus, ServiceDescription, ServiceType, ServiceClient, ServiceClientType, Service, Endpoint
 from xrdsst.main import XRDSSTTest
 from xrdsst.rest.rest import ApiException
```
bertvi, line 13: 1. tests/unit/test_endpoint.py: line 97-98 "with mock.patch('xrdsst.api.services_api.ServicesApi.add_endpoint',"
   This should probably return Endpoint() not ServiceDescription()?
 
2. tests/unit/test_endpoint.py: lines 179-181 and 214-216 "                        with mock.patch(
                                'xrdsst.api.endpoints_api.EndpointsApi.add_endpoint_service_clients',
                                return_value=EndpointTestData.add_description_response):"
                                
   these should probably return [ServiceClient()] not ServiceDescription?
  
  
3. tests/unit/test_endpoint.py lines 257-259 "                        with mock.patch(
                                'xrdsst.api.endpoints_api.EndpointsApi.add_endpoint_service_clients',
                                return_value=EndpointTestData.add_description_response):"
    this should probably return [ServiceClient()] not ServiceDescription()?

xrdsst/controllers/endpoint.py
```diff
@@ -4,19 +4,18 @@
...
-
 class EndpointController(BaseController):
     class Meta:
         label = 'endpoint'
         stacked_on = 'base'
         stacked_type = 'nested'
         description = texts['endpoint.controller.description']
 
-    @ex(label='add-endpoints', help="Add endpoints", arguments=[])
+    @ex(help="Add endpoints", arguments=[])
```
bertvi, line 18: If it is not too difficult, please change this to "add" like in ClientController, because "xrdsst endpoints add" would probably look better than "xrdsst endpoints add-endpoint":

    @ex(help="Add client subsystem", arguments=[])
    def add(self):

xrdsst/controllers/endpoint.py
```diff
@@ -56,20 +72,82 @@ def remote_add_service_endpoints(self, ss_api_config, security_server_conf, clie
...
+
+                            except ApiException as find_err:
+                                BaseController.log_api_error('ClientsApi->get_client_service_description', find_err)
+                except ApiException as find_err:
+                    BaseController.log_api_error('ClientsApi->get_client_service_description', find_err)
+        except ApiException as find_err:
+            BaseController.log_api_error('ClientsApi->find_clients', find_err)
+
+
+    def getServicesDescription(self, config):
```
bertvi: method "getServicesDescription()" would be ok for Java, but Python uses snake-case, so probably get_services_description()" would be better.(https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

bertvi (changes requested):

bertvi (approved):

PR #61, opened 2021-05-14
SSTOOLKIT-79

bertvi:
Added sub-command "update" to "xrdsst client" to update subsystem parameters, especially "connection_type"

### DIFF
--- a/xrdsst/controllers/client.py
+++ b/xrdsst/controllers/client.py
@@ -146,17 +146,8 @@ class ClientController(BaseController):
                     return
 
                 try:
-                    new_client = Client(id=client.id,
-                                        instance_id=client.instance_id,
-                                        member_name=client.member_name,
-                                        member_class=client.member_class,
-                                        member_code=client.member_code,
-                                        subsystem_code=client.subsystem_code,
-                                        owner=client.owner,
-                                        has_valid_local_sign_cert=client.has_valid_local_sign_cert,
-                                        connection_type=convert_swagger_enum(ConnectionType, client_conf['connection_type']),
-                                        status=client.status)
-                    response = clients_api.update_client(client.id, body=new_client)
+                    client.connection_type = convert_swagger_enum(ConnectionType, client_conf['connection_type'])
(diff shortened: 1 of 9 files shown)


xrdsst/controllers/client.py
```diff
@@ -108,6 +134,29 @@ def remote_register_client(self, ss_api_config, security_server_conf, client_con
...
+                        security_server_conf['name'] + ": " + self.partial_client_id(client_conf) + " not added/registered yet."
+                    )
+                    return
+
+                try:
+                    client.conn_type = convert_swagger_enum(ConnectionType, client_conf['connection_type'])
+                    client.member_class = client_conf['member_class']
+                    client.member_code = client_conf['member_code']
+                    client.subsystem_code = client_conf['subsystem_code']
+                    clients_api.update_client(id=client.id)
```
bertofl90: You are not sending the clients object in the clients_api request, apart form that, can we update the member class and member code with this method? could not this cause an error?

bertofl90 (changes requested):

xrdsst/controllers/client.py
bertvi: The update_client() method takes client id as argument: 

    def update_client(self, id, **kwargs):  # noqa: E501
        """update security server client information  # noqa: E501

        <h3>Administrator updates the client information.</h3>  # noqa: E501
        This method makes a synchronous HTTP request by default. To make an
        asynchronous HTTP request, please pass async_req=True
        >>> thread = api.update_client(id, async_req=True)
        >>> result = thread.get()

        :param async_req bool
        :param str id: id of the client (required)

I removed updating of other parameters, only connection_type remains as changing of other parameters is not possible and will  cause an error. In order to change other parameters, a new client must be added instead.

bertofl90: But how it can't update the properties if you don't send the client in the body. As is the code right now it is possible to change other parameters, for example putting different values for the member class in the configuration file

bertofl90 (approved):

PR #62, opened 2021-05-17
SSTOOLKIT-78

bertofl90:


docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -642,3 +647,43 @@ to use nightly backups that are kept at security server to revert to earlier sta
 Overview of existing automatic backups is accessible from web administration console
 of the security server, in the "Settings" menu. More information about functionality
 can be found in [UG-SS](#Ref_SS-UG).
+
+## 6 Multitenancy
+It's possible to add another members and subsystem to a security server using the toolkit.
+For doing that we need to add the members and subsystems in the clients section of the configuration
+file. 
+For adding a new member we must delete the properties 'service_descriptions' and 'subsystem_code'.
```
bertvi, line 749: Why do you need to remove "service_descriptions"? When client is added, these should not affect anything as they are ignored, only used when service descriptions are added.

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -642,3 +647,43 @@ to use nightly backups that are kept at security server to revert to earlier sta
...
+For example if we want to add the member 'COM/12345/COMPANY', and the subsystem 'COM/12345/COMPANY/SUB' we must fill the
+configuration file like this:
+
+```
+[...]
+  clients:
+    - member_class: <MEMBER_CLASS>
```
bertvi: Why do you keep placeholders here, but later on use real example values? I think you should stick to the same style, either use only placeholders or substitute all with real example values.

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -642,3 +647,43 @@ to use nightly backups that are kept at security server to revert to earlier sta
...
+      connection_type: HTTP
+      service_descriptions:	
+      	[...]	    
+```
+
+We can add new members/subsystems with the command:
+```
+xrdsst client add-client
+```
+The command will also create a SIGN CSRS certificate for this new member/subsystem. We can download this CSRS with the command:
```
bertvi: Is it necessary to mention the CSRS here, because when adding clients before, we also needed to run "cert download-csrs" in any case? CSRS are needed when importing certificates, these are not really related to adding clients.

xrdsst/controllers/endpoint.py
```diff
@@ -45,8 +46,9 @@ def add_service_endpoints(self, config):
         ss_api_conf_tuple = list(zip(config["security_server"], map(lambda ss: self.create_api_config(ss, config), config["security_server"])))
 
         for service_description_dic in self.get_services_description(config):
-            for endpoint_conf in service_description_dic["service_description"]["endpoints"]:
-                self.remote_add_service_endpoints(service_description_dic["ss_api_config"], service_description_dic["security_server"], service_description_dic["client"], service_description_dic["service_description"], endpoint_conf)
+            if "endpoints" in service_description_dic["service_description"]:
```
bertvi, line 51: Why is this needed?

xrdsst/core/validator.py
```diff
@@ -175,10 +175,10 @@ def validate_config_client_add_or_register(ss_config, operation, errors):
                 ss_config_client_slice, operation, errors
             )
 
-            require_fill(
-                ConfKeysSecServerClients.CONF_KEY_SS_CLIENT_SUBSYSTEM_CODE,
-                ss_config_client_slice, operation, errors
-            )
+            # require_fill(
```
bertvi: We should not keep code that is commented out. Maybe you should create some kind of conditional requirement here, so that when member_name is filled, subsystem_code is not required and vice versa, if subsystem_code is filled, member_name is not required, because for adding subsystems, subsystem_code is required and when adding members, member_name is required, if both are not required, it will not work

bertvi (changes requested):
1. When running end to end tests, I got an exception: "FAILED tests/end_to_end/tests.py::EndToEndTest::test_run_configuration - KeyError: 'service_descriptions'".

2. Maybe it would make more sense to change the current "client add" method to something like "add-subsystem" and create a new method "add-member" for adding member? That way E2E tests could also be properly run, one test for adding member and another one for adding a subsystem.

docs/xroad_security_server_toolkit_user_guide.md
bertofl90: I have commented on it because it can create confusion and potentially failures since members cannot have service_descriptions

xrdsst/controllers/endpoint.py
bertofl90: I think this is not related with the task but I found a bug here testing it all, we must check that the service description have endpoints, if not could fail if one of the service descriptions in the configuration doesn't have endpoints configured

docs/xroad_security_server_toolkit_user_guide.md
bertofl90: If you add a new member through the interface you will see that it adds also a new SIGN key and CSRS for this member, I try to recreate this behaviour to and the command add-clients for new members also creates the CSRS, and if you want to got it for signing you need to run the download-csrs command.

bertvi (approved):

PR #87, opened 2021-06-25
SSTOOLKIT-102

bertvi:


xrdsst/controllers/service.py
```diff
@@ -327,6 +374,32 @@ def remote_update_service_parameter(ss_api_config, service_description_conf, ser
         except ApiException as err:
             BaseController.log_api_error('ServicesApi->update_service', err)
 
+    def remote_list_service_descriptions(self, ss_api_config, security_server, client):
+        clients_api = ClientsApi(ApiClient(ss_api_config))
+        try:
+            service_descriptions_list = []
+            render_data = []
+            client_ids = client.split(',')
```
bertofl90: i've created in another branch a this function in util.py:
def parse_argument_list(arg):
    list = arg.split(',')
    return list
    
It's just s simple split but maybe it's better to have it centralize in one single place, if not, if in the future we want to change the character separator we will need to find it through all the code

xrdsst/controllers/service.py
```diff
@@ -327,6 +374,32 @@ def remote_update_service_parameter(ss_api_config, service_description_conf, ser
...
+            for client_id in client_ids:
+                service_descriptions = clients_api.get_client_service_descriptions(id=client_id)
+                for service_description in service_descriptions:
+                    service_descriptions_list.append({'security_server': security_server["name"],
+                                                      'client_id': client_id,
+                                                      'description_id': service_description.id,
+                                                      'url': service_description.url,
+                                                      'type': service_description.type,
+                                                      'disabled': service_description.disabled,
+                                                      'services': len(service_description.services)})
```
bertofl90: Maybe here check if the services exists just in case, something like: len(service_description.services) if service_description.services else 0

xrdsst/controllers/service.py
bertvi: I added the same functionality and used it

xrdsst/controllers/service.py
bertvi: Added the check

PR #92, opened 2021-07-02
SSTOOLKIT-105

bertvi:


bertofl90 (commented):
Send the security server as parameter it's probably required to distinguish clients between security servers

xrdsst/controllers/service.py
```diff
@@ -163,6 +164,57 @@ def list_services(self):
...
+
+    @ex(help="Update service descriptions", arguments=[(['--ss'], {'help': 'Security server name', 'dest': 'ss'}),
+                                                       (['--client'], {'help': 'Client id', 'dest': 'client'}),
+                                                       (['--description'], {'help': 'Service description id', 'dest': 'description'}),
+                                                       (['--code'], {'help': 'REST service code', 'dest': 'code'}),
+                                                       (['--url'], {'help': 'Service description url', 'dest': 'url'})])
+    def update_descriptions(self):
+        active_config = self.load_config()
+
+        if self.app.pargs.ss is None:
```
bertofl90, line 195: Maybe here and in the other commands is better to remove the return on each "if" and just put an error flag or something, in that case, we could show the full list of missing parameters instead of one by one

tests/unit/test_service.py
```diff
@@ -633,4 +651,92 @@ def test_service_list_services_fail_description_missing(self):
...
+                    service_controller.update_descriptions()
+
+                    out, err = self.capsys.readouterr()
+                    assert out.count("updated successfully") > 0
+
+                    with self.capsys.disabled():
+                        sys.stdout.write(out)
+                        sys.stderr.write(err)
+
+    def test_service_update_descriptions_fail_client_missing(self):
```
bertofl90, line 708: You are not reading the output in this unit tests

bertofl90 (approved):

PR #97, opened 2021-07-07
SSTOOLKIT-108

bertvi:


docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -1018,7 +1021,36 @@ xrdsst service disable-descriptions --ss <SECURITY_SERVER_NAME> --client <CLIENT
 * <SERVICE_DESCRIPTION_ID> id of the service description, e.g. 123, multiple values can also be given, separated by comma, e.g. 123,456
 * <NOTICE> disabling notice, e.g. "Not used"
 
-##### 4.2.7.11 Service apply
+##### 4.2.7.11 Service list access rights for services
+
+* Access rights: XROAD_SYSTEM_ADMINISTRATOR and XROAD_SERVICE_ADMINISTRATOR
```
bertofl90: Please review this permissions, i don't think that a read action requires this two roles

xrdsst/controllers/service.py
```diff
@@ -66,6 +67,38 @@ def as_object(service):
...
+    def as_object(service):
+        return {
+            'security_server': service.get('security_server'),
+            'client_id': service.get('client_id'),
+            'description_id': service.get('description_id'),
+            'service_id': service.get('service_id'),
+            'service_client_id': service.get('service_client_id'),
+            'local_group': service.get('local_group'),
+            'name': service.get('name'),
+            'rights_given': service.get('rights_given'),
```
bertofl90, line 95: This date should be formatted, check the example in certificates list

xrdsst/controllers/service.py
```diff
@@ -66,6 +67,38 @@ def as_object(service):
         }
 
 
+class ServiceAccessListMapper:
+    @staticmethod
+    def headers():
+        return ['SS', 'CLIENT', 'DESCRIPTION', 'SERVICE', 'SERVICE_CLIENT', 'LOCAL_GROUP', 'NAME', 'RIGHTS_GIVEN', 'TYPE']
```
bertofl90: The table is a little confusing, too much information. We could remove the client and description columns since we are sending them as parameters, also I would put the local group name in the column of service clients since the Id of the local group doesn't provide much information here.

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -1018,7 +1021,36 @@ xrdsst service disable-descriptions --ss <SECURITY_SERVER_NAME> --client <CLIENT
...
+Deleting access rights for services for client's service descriptions can be done with:
+```
+xrdsst service delete-access --ss <SECURITY_SERVER_NAME> --client <CLIENT_ID> --description <SERVICE_DESCRIPTION_ID> --service <SERVICE_ID> --sclient <SERVICE_CLIENT_ID>
+```
+* <SECURITY_SERVER_NAME> name of the security server, e.g. ss1
+* <CLIENT_ID> id of the client, e.g. DEV:GOV:1234:TEST
+* <SERVICE_DESCRIPTION_ID> id of the service description, e.g. 123, multiple values can also be given, separated by comma, e.g. 123,456
+* <SERVICE_ID> id of the service, e.g. DEV:GOV:1234:TEST:Petstore, multiple values can also be given, separated by comma, e.g. DEV:GOV:1234:TEST:Petstore,DEV:GOV:1234:TEST:Petstore2
+* <SERVICE_CLIENT_ID> id of the service description, e.g. DEV:security-server-owners, multiple values can also be given, separated by comma, e.g. DEV:GOV:1234:TEST,DEV:security-server-owners
+
```
bertofl90, line 1052: Here we could describe how the command works when multiple clients, services, or descriptions are entered. Because it is not entirely clear how it works, from what I see it goes through everything and if it does not find the sclient it skipped it without putting any message

xrdsst/controllers/service.py
```diff
@@ -260,6 +293,52 @@ def disable_descriptions(self):
...
+            missing_parameters.append('description')
+        if self.app.pargs.service is None:
+            missing_parameters.append('service')
+        if self.app.pargs.sclient is None:
+            missing_parameters.append('sclient')
+        if len(missing_parameters) > 0:
+            BaseController.log_info('The following parameters missing for deleting access rights for services: %s' % missing_parameters)
+            return
+
+        self.delete_access_rights(active_config,
```
bertofl90, line 349: Also i see in your commands, that normally parse the parameters in the "remote" method, but I think it's better if we parse the parameters here so we don't need to go deeper in the code to check if the parameter could be multiple or if we must send a list or a string separated by ','

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Updated

xrdsst/controllers/service.py
bertvi: Updated

xrdsst/controllers/service.py
bertvi: Deleted local_group, but "ss", "client" and "description" should be kept because clients and descriptions with same values could exist on multiple security servers and it will be confusing when we would list duplicates without knowing which server or client these belong to

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Refactored the code and updated the documentation, all operations now consist only a single parameter allowing multiple values so it should not be confusing anymore

xrdsst/controllers/service.py
bertvi: Updated

PR #102, opened 2021-07-15
SSTOOLKIT-116

bertofl90:


tests/end_to_end/tests.py
```diff
@@ -1290,6 +1288,17 @@ def step_client_delete(self):
                     found_client = get_client(self.config, client, ssn)
                     assert len(found_client) == 0
 
+    def step_make_owner(self):
+        member = 'DEV:COM:222'
+        with XRDSSTTest() as app:
+            client_controller = ClientController()
+            client_controller.app = app
+            configuration = client_controller.create_api_config(self.config["security_server"][0], self.config)
```
bertvi: It seems you are performing the operation for all security servers, but creating the api key only for the first one, it will probably not work that way

tests/end_to_end/tests.py
```diff
@@ -1427,6 +1436,7 @@ def test_run_configuration(self):
 
         RenewCertificate(self).test_run_configuration()
         LocalGroupTest(self).test_run_configuration()
+        self.step_make_owner()
```
bertvi: It would be less confusing if the step name would also include "client", e.g step_client_make_owner

tests/integration/test_xrdsst.py
```diff
@@ -1012,6 +1012,15 @@ def step_client_unregister(self):
                     assert len(found_client) > 0
                     assert found_client[0]["status"] == ClientStatus.DELETION_IN_PROGRESS
 
+    def step_make_owner(self):
+        member = 'DEV:COM:222'
+        with XRDSSTTest() as app:
+            client_controller = ClientController()
+            client_controller.app = app
+            configuration = client_controller.create_api_config(self.config["security_server"][0], self.config)
```
bertvi: It seems you are performing the operation for all security servers, but creating the api key only for the first one, it will probably not work that way

tests/integration/test_xrdsst.py
```diff
@@ -1144,6 +1153,7 @@ def test_run_configuration(self):
 
         LocalGroupTest(self).test_run_configuration()
         RenewCertificate(self).test_run_configuration()
+        self.step_make_owner()
```
bertvi: It would be less confusing if the step name would also include "client", e.g step_client_make_owner

tests/unit/test_client.py
```diff
@@ -30,7 +30,7 @@ class ClientTestData:
     )
 
     update_response = Client(
-        id='DEV:GOV:9876:SUB1',
+        id='DEV:GOV:9876',
```
bertvi: Why is this needed?

xrdsst/controllers/client.py
```diff
@@ -118,6 +119,25 @@ def delete(self):
...
+    def make_owner(self):
+        active_config = self.load_config()
+        missing_parameters = []
+        if self.app.pargs.ss is None:
+            missing_parameters.append('ss')
+        if self.app.pargs.member is None:
+            missing_parameters.append('member')
+        if len(missing_parameters) > 0:
+            BaseController.log_info(
+                'The following parameters missing for deleting clients: %s' % missing_parameters)
```
bertvi: Copy-paste error

xrdsst/core/util.py
```diff
@@ -8,19 +8,23 @@
 
 
 def get_admin_credentials(security_server, config):
-    admin_credentials_env_variable = security_server["admin_credentials"] if security_server.get("admin_credentials", "") else config["admin_credentials"]
+    admin_credentials_env_variable = security_server["admin_credentials"] if security_server.get("admin_credentials",
```
bertvi: Why have you changed this file?

bertvi (changes requested):

tests/unit/test_client.py
bertofl90: i have deleted it, it's not used in the class

xrdsst/core/util.py
bertofl90: i don't remember change this, probably it was without notice

bertvi (approved):

PR #104, opened 2021-07-16
SSTOOLKIT-125

bertvi:


config/xrdsst.yml
```diff
@@ -1,60 +1,169 @@
-admin_credentials: <SECURITY_SERVER_CREDENTIALS_OS_ENV_VAR_NAME>
+admin_credentials: TOOLKIT_ADMIN_CREDENTIALS
```
bertofl90: please revert configuration changes

docs/xroad_security_server_toolkit_user_guide.md
```diff
@@ -1353,6 +1358,28 @@ xrdsst backup restore --ss <SECURITY_SERVER_NAME> --file <BACKUP_FILENAME>
 
 * <SECURITY_SERVER_NAME> name of the security server, e.g. ss1
 * <BACKUP_FILENAME> file name of the backup to be restored from, e.g. conf_backup_20210713-161054.tar
+
+#### 4.2.12 Diagnostics management
+
```
bertofl90, line 1364: In other tasks that are listed, we put the table with a brief description of the properties that are shown, it would be good to add them for a bit of consistency

xrdsst/controllers/diagnostics.py
```diff
@@ -0,0 +1,207 @@
+from cement import ex
```
bertofl90, line 1: I think this whole code could be refactor, because it's practically the same for all diagnostics, look what I did for example in the cert controller with the class cert operation, you could have more or less the same, one object with the API method and other call parse_response or print_response or something like that.

xrdsst/controllers/diagnostics.py
```diff
@@ -0,0 +1,207 @@
...
+
+from xrdsst.api import DiagnosticsApi
```
bertofl90, line 3: Also, it could be helpful another command for printing all the diagnostics instead of going one by one

config/xrdsst.yml
bertvi: Done

docs/xroad_security_server_toolkit_user_guide.md
bertvi: Added example and description of the output

xrdsst/controllers/diagnostics.py
bertvi: Refactored the code and removed 3 similar methods

xrdsst/controllers/diagnostics.py
bertvi: Added "diagnostics all" command, which calls all the methods and prints 3 tables

bertofl90 (approved):

PR #119, opened 2021-08-19
fix finnish profile name

bertofl90:


bertvi (changes requested):
Do a search of "FIVRK" in the project, that name is still used in the User Guide and also in tests and names of some classes, I suggest change to FI also in those places so that "FIVRK" is not mentioned anymore.

bertvi (approved):

PR #121, opened 2022-10-10
Changes related to the handover from DVV to NIIS

raits:
This PR implement changes required for handing the project over to NIIS. The version info still needs to be updated once we have our first official release, but this updates the following:

- License text
- setup.py with NIIS information
- Adds the SkKlass3 profile for Estonia
- Changes the default profile to Ejbca
- Sets the docker environment to use version 3.6.8 of the Python docker container rather than just 3 (latest containers fail to compile the project)
- Updated unit-tests related to the profiles


setup.py
```diff
@@ -20,8 +20,8 @@
     long_description_content_type='text/markdown',
     classifiers=[],
     install_requires=required,
-    author='Finnish Digital Agency',
-    author_email='info@dvv.fi',
+    author='Nordic Institute for Interoperability Solutions (NIIS)',
+    author_email='info@niis.org',
```
petkivim: Let's use `x-road@niis.org` here.

petkivim (approved):

setup.py
raits: Updated

PR #125, opened 2024-03-13
update `jq` version to 1.6.0

KulykDmytro:
fix issue with installation `xrdsst` onto images newer than `ubuntu:20.04` or bare `python` ones

```bash
jq.c:196:12: fatal error: longintrepr.h: No such file or directory
  196 |   #include "longintrepr.h"
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ bump2version~=1.0.1
 gitpython~=3.1.11
 docker~=4.1.0
 yq~=2.11.1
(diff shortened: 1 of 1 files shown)


raits:
Hello, thank you for your contribution @KulykDmytro! Since the changes are just to bump the version of a dependency and don’t introduce fundamental changes to the code or documentation we can accept these updates without a signed CLA.

We will try to review and merge the request soon.

